PR #72, opened 2024-10-29
Cleanups

christianparpart:


src/Lightweight/DataBinder/UnicodeConverter.hpp
```diff
@@ -88,6 +88,14 @@ LIGHTWEIGHT_API std::u8string ToUtf8(std::u32string_view u32InputString);
 // Converts from UTF-16 to UTF-8.
 LIGHTWEIGHT_API std::u8string ToUtf8(std::u16string_view u16InputString);
 
+// Converts from UTF-16 (as wchar_t) to UTF-8.
+template <typename T>
+    requires(std::is_same_v<T, wchar_t> && sizeof(wchar_t) == 2)
```
Yaraslaut, line 93: just a small comment, `std::same_as` is a concept, while `std::is_same` is a type from, maybe better to use concept ?

PR #74, opened 2024-11-01
Add benchmarks

Yaraslaut:
Add benchmarks

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -48,3 +48,4 @@ add_subdirectory(src/tools)
 
 enable_testing()
 add_subdirectory(src/tests)
+add_subdirectory(src/benchmark)
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ build/
 vcpkg_installed/
 CMakeSettings.json
 compile_commands.json
+*.sqlite
--- a/src/benchmark/CMakeLists.txt
+++ b/src/benchmark/CMakeLists.txt
@@ -1,0 +1,7 @@
+
+option(LIGHWEIGHT_BENCHMARK "Enable lightweight benchmark" ON)
+
+# Enable benchmark
+add_executable(LightweightBenchmark benchmark.cpp)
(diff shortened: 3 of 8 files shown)


CMakeLists.txt
```diff
@@ -48,3 +48,4 @@ add_subdirectory(src/tools)
 
 enable_testing()
 add_subdirectory(src/tests)
+add_subdirectory(benchmark)
```
christianparpart, line 51: wouldn't it make more sense in `src/benchmark`? :thinking:

benchmark/benchmark.cpp
```diff
@@ -0,0 +1,234 @@
+#include <print>
```
christianparpart, line 1: SPDX-header missing

src/tests/Utils.hpp
```diff
@@ -210,7 +210,7 @@ class SqlTestFixture
 
     virtual ~SqlTestFixture() = default;
 
-  private:
+    private:
```
christianparpart, line 213: changed by accident?

benchmark/benchmark.cpp
```diff
@@ -0,0 +1,234 @@
...
+std::string SanitizePwd(std::string_view input) {
+  std::regex const pwdRegex{
+      R"(PWD=.*?;)",
+      std::regex_constants::ECMAScript | std::regex_constants::icase,
+  };
+  std::stringstream outputString;
+  std::regex_replace(std::ostreambuf_iterator<char>{outputString},
+                     input.begin(), input.end(), pwdRegex, "Pwd=***;");
+  return outputString.str();
+}
```
christianparpart, line 22: looks like a little bit of copy'n'paste from src/tests/Utils.hpp - could we probably avoid this? :thinking: 

Also, the indentation width seems to be `2` rather than `4` here. Is this intentionally, or just by accident, as `src/.clang-format` did not apply for `benchmark/**`, because it's not living in a descendant root directory?

christianparpart (approved):
Thank you very much, @Yaraslaut :)

PR #84, opened 2024-11-07
SqlNullValue fix for MS SQL server and improve overall NULL value support

### LINKED ISSUE #83: Add test case(s) for writing NULL to nullable columns
to make sure it works for all column types we support

christianparpart:
SqlNullType: when INSERT'ing this to a nullable INT column, MS SQL se…rver was raising an error, because of an invalid precision. This fixes it.

closes #83


Yaraslaut:
Is it possible to add unit test for this case? this is a very specific behavior from MS SQL side

christianparpart:
haha. thx for asking, @Yaraslaut - this is basically https://github.com/christianparpart/Lightweight/issues/83 then :)

I was just wondering if i do it right here or in a follow-up PR, because it was actually kind-of time-pressing in the company, as a developer was running into it and was wondering why his prepared INSERT statement was throwing. It took me a while to figure out as well. We'll definitely add some test(s) for all data types (at best).

Yaraslaut:
Ah, alright then, we just need to make sure that this will be covered :)

PR #87, opened 2024-11-08
Add new column type `SqlNumeric<P, S>`

christianparpart:
Mind, `NUMERIC(P, S)` does not seem to be supported by SQLite3, too bad.

We could probably emulate that with REAL at best.

@FelixTheC this is used for things like curreny, e.g. prices, payment, etc...

### DIFF
--- a/src/Lightweight/DataBinder/SqlNumeric.hpp
+++ b/src/Lightweight/DataBinder/SqlNumeric.hpp
@@ -1,0 +1,151 @@
+// SPDX-License-Identifier: Apache-2.0
+
+#pragma once
+
+#include "Primitives.hpp"
+
+#include <cmath>
+#include <compare>
+#include <concepts>
+#include <cstring>
+
+// clang-format off
+#if defined(__SIZEOF_INT128__)
+    #define LIGHTWEIGHT_INT128_T __int128_t
+    static_assert(sizeof(__int128_t) == SQL_MAX_NUMERIC_LEN);
+#endif
+// clang-format on
+
+// Represents a fixed-point number with a given precision and scale.
+//
+// Precision is *exactly* the total number of digits in the number,
+// including the digits after the decimal point.
+//
+// Scale is the number of digits after the decimal point.
+template <std::size_t ThePrecision, std::size_t TheScale>
+struct SqlNumeric
+{
+    // Number of total digits
+    static constexpr auto Precision = ThePrecision;
+
+    // Number of digits after the decimal point
+    static constexpr auto Scale = TheScale;
+
+    static_assert(TheScale < SQL_MAX_NUMERIC_LEN);
+    static_assert(Scale <= Precision);
+
+    // The value is stored as a string to avoid floating point precision issues.
+    SQL_NUMERIC_STRUCT sqlValue {};
+
+    constexpr SqlNumeric() noexcept = default;
+    constexpr SqlNumeric(SqlNumeric&&) noexcept = default;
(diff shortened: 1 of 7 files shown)


src/Lightweight/DataBinder/SqlNumeric.hpp
```diff
@@ -0,0 +1,141 @@
...
+    {
+        auto const num = std::atoll((char const*) sqlValue.val);
+        auto const sign = sqlValue.sign ? 1 : -1;
+        auto const p = std::pow(10, sqlValue.scale);
+        return sign * num / p;
+    }
+
+    [[nodiscard]] LIGHTWEIGHT_FORCE_INLINE std::string ToString() const
+    {
+        return std::format("{:.{}f}", ToFloat(), 2); //(unsigned) sqlValue.precision);
```
FelixTheC, line 88: this doesn't look like a usefull comment `//(unsigned) sqlValue.precision);`

src/tests/DataBinderTests.cpp
```diff
@@ -739,4 +751,53 @@ TEST_CASE_METHOD(SqlTestFixture, "SqlDataBinder: SqlGuid", "[SqlDataBinder],[Sql
...
+    REQUIRE(actual.has_value());
+    CHECK(*actual == expectedValue);
+
+    // Fetch and check GetColumn for GUID
+    {
+        stmt.ExecuteDirect("SELECT Value FROM Test");
+        REQUIRE(stmt.FetchRow());
+        auto const actualValue = stmt.GetColumn<SqlNumeric<10, 2>>(1);
+        CHECK(actualValue == expectedValue);
+    }
```
FelixTheC, line 783: In catch2 you can use [SECTION](https://github.com/catchorg/catch2/blob/devel/docs/tutorial.md#test-cases-and-sections) if you want to have also an information in the test output
```suggestion
    
    SECTION("Fetch and check GetColumn for GUID") {
        stmt.ExecuteDirect("SELECT Value FROM Test");
        REQUIRE(stmt.FetchRow());
        auto const actualValue = stmt.GetColumn<SqlNumeric<10, 2>>(1);
        CHECK(actualValue == expectedValue);
    }
```

src/tests/DataBinderTests.cpp
```diff
@@ -739,4 +751,53 @@ TEST_CASE_METHOD(SqlTestFixture, "SqlDataBinder: SqlGuid", "[SqlDataBinder],[Sql
...
+
+    // Bind output column, fetch, and check result in output column for GUID
+    {
+        stmt.Prepare("SELECT Value FROM Test");
+        SqlNumeric<10, 2> actualValue;
+        stmt.BindOutputColumns(&actualValue);
+        stmt.Execute();
+        REQUIRE(stmt.FetchRow());
+        CHECK(actualValue == expectedValue);
+    }
```
FelixTheC, line 793:
```suggestion
    
    SECTION("Bind output column, fetch, and check result in output column for GUID") {
        stmt.Prepare("SELECT Value FROM Test");
        SqlNumeric<10, 2> actualValue;
        stmt.BindOutputColumns(&actualValue);
        stmt.Execute();
        REQUIRE(stmt.FetchRow());
        CHECK(actualValue == expectedValue);
    }
```

src/Lightweight/DataBinder/SqlNumeric.hpp
christianparpart: oh nice. that was a hnag-over. THX!

PR #88, opened 2024-11-11
Unify DataBinder<> tests and fix some findings along the way

christianparpart:
- rework data binder tests to templated parametrized a test case
- and add checks for types `float` and `double`
- make some more SQL column types `constexpr`
- SqlFixedString: refactor null terminator handling
- SqlNumeric: Fix sending/retrieval for Oracle SQL server
- SqlTrimmedString: Fix right-trimming NULL values
- SqlStatement: Fix GetColumn() handling of nullable T's
- SqlNumeric: Add **explicit** native type cast operator overload
- more minors
- tests: perform sql trace logging through catch2's UNSCOPED_INFO

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -250,7 +250,7 @@ jobs:
           echo "ODBC_CONNECTION_STRING=${{ steps.setup.outputs.ODBC_CONNECTION_STRING }}"
           ldd /home/runner/oracle/instantclient_21_3/libsqora.so.21.1
       - name: "Run SQL Core tests"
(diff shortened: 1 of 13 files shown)


christianparpart:
@FelixTheC this is FYI, but still worth a look, especially into DataBinderTests.cpp (bottom), because I've fully unified the way of testing column data type binding, such that every data type is tested exactly the same way. If you find something, I am still appreciating it very much and will adapt in a follow-up PR. - this one here I need for the reflection ORM PR, so I'd prefer to merge rather early

PR #89, opened 2024-11-11
minor code cleanup

christianparpart:


src/Lightweight/DataBinder/SqlFixedString.hpp
```diff
@@ -193,6 +203,21 @@ class SqlFixedString
     }
 };
 
+template <typename>
+struct IsSqlFixedStringType
```
Yaraslaut, line 207: I think better to use std::true(false)_type

christianparpart: It is. But I am wasted. Will look at this breakout PR again after some recharging :)

PR #105, opened 2024-11-25
Use reflection to select all fields in the structure

Yaraslaut:
Implemented `Field()` function

### DIFF
--- a/src/tests/QueryBuilderTests.cpp
+++ b/src/tests/QueryBuilderTests.cpp
@@ -1,6 +1,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 #include "Utils.hpp"
+#include <Lightweight/DataMapper/DataMapper.hpp>
 
 #include <catch2/catch_session.hpp>
 #include <catch2/catch_test_macros.hpp>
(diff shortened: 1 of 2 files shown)


src/tests/QueryBuilderTests.cpp
```diff
@@ -100,13 +100,27 @@ TEST_CASE_METHOD(SqlTestFixture, "SqlQueryBuilder.Select.Range", "[SqlQueryBuild
...
 
+struct Users
+{
+    std::string name;
+    std::string address;
+};
+
+TEST_CASE_METHOD(SqlTestFixture, "SqlQueryBuilder.Fields", "[SqlQueryBuilder]")
+{
+    checkSqlQueryBuilder([](SqlQueryBuilder& q) { return q.FromTable("Users").Select().Fields<Users>().First(); },
```
christianparpart, line 111: very nice. thanks.

I think I failed last time, because I did it with data mapper in mind. The example you have here should definitely continue working. But do you think we can extend this PR to also make it work with data mapper aware structs?

Such as:

```cpp
struct Users
{
    Field<std::string> name;
    Field<std::optional<std::string>> address; // Also BelongsTo<> should work
};
```

src/tests/QueryBuilderTests.cpp
```diff
@@ -100,13 +100,27 @@ TEST_CASE_METHOD(SqlTestFixture, "SqlQueryBuilder.Select.Range", "[SqlQueryBuild
...
+                         });
+}
+
 TEST_CASE_METHOD(SqlTestFixture, "SqlQueryBuilder.Delete", "[SqlQueryBuilder]")
 {
     checkSqlQueryBuilder(
         [](SqlQueryBuilder& q) { return q.FromTable("That").Delete().Where("foo", 42).Where("bar", "baz"); },
         QueryExpectations::All(R"(DELETE FROM "That" WHERE "foo" = 42 AND "bar" = 'baz')"));
 }
-
```
christianparpart, line 109: I guess this line was deleted by accident. Can we get it back? (I know, sounds trivial, but it might make sense to keep a single empty line between function declarations)

Yaraslaut: Yes, deleted by accident :)

src/tests/QueryBuilderTests.cpp
Yaraslaut: Added more tests, to conver usage of `Field` and `BelongsTo`

christianparpart (approved):
thanks @Yaraslaut. This will be definitely helpful.

Not sure why this passed clang-format (test file looks off). But we can address this any time later. :)

PR #107, opened 2024-11-25
Add input_iterator API wrapper over sql statement

Yaraslaut:
PR implements following use case:
```
struct movies final
{
    Field<int, PrimaryKey::Manual> movie_id;
    Field<SqlTrimmedFixedString<25>> movie_title;
};
```

```
auto stmt = SqlStatement {};
for(auto&& val: iterator<movies>(stmt, "movies"))
    std::println("{}{}", val.movie_id.Value(),val.movie_title.Value());
```

### DIFF
--- a/src/Lightweight/SqlStatement.hpp
+++ b/src/Lightweight/SqlStatement.hpp
@@ -323,6 +323,64 @@ class [[nodiscard]] SqlResultCursor
     SqlStatement* m_stmt;
 };
 
+// input iterator
+template <typename T>
+class SqlRowIterator
+{
+  public:
+    using difference_type = bool;
+    using value_type = T;
+
+    SqlRowIterator(SqlStatement& stmt):
+        _stmt(&stmt)
+    {
+    }
+
+    SqlRowIterator operator++()
+    {
+        _is_end = _stmt->FetchRow();
+        return *this;
+    }
+
+    value_type operator*()
+    {
+        auto res = T {};
+
+        Reflection::EnumerateMembers(res, [this]<size_t I>(auto&& value) {
+            auto tmp = _stmt->GetColumn<typename Reflection::MemberTypeOf<I, value_type>::ValueType>(I + 1);
+            value = tmp;
+        });
+
+        return res;
+    }
+
+    SqlRowIterator begin()
+    {
+        auto query =
+            _stmt->Query(detail::RecordTableName<value_type>::Value).Select().template Fields<value_type>().All();
+        _stmt->Prepare(query);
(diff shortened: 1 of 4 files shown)


src/Lightweight/SqlStatement.hpp
```diff
@@ -321,6 +321,67 @@ class [[nodiscard]] SqlResultCursor
...
+template <typename T>
+class iterator
+{
+  public:
+    using difference_type = bool;
+    using value_type = T;
+
+    iterator(SqlStatement& stmt, std::string_view table_name):
+        m_stmt(&stmt),
+        m_table_name(table_name)
```
christianparpart, line 334:
```suggestion
        m_table_name(RecordTableName<T>)
```
If `T` is supposed to be the record struct, then we do not need to pass this string_view to the constructor.

src/Lightweight/SqlStatement.hpp
```diff
@@ -321,6 +321,67 @@ class [[nodiscard]] SqlResultCursor
...
+    {
+    }
+
+    iterator operator++()
+    {
+        if (!m_stmt->FetchRow())
+            m_is_end = true;
```
christianparpart, line 341:
```suggestion
        m_is_end = !m_stmt->FetchRow();
```
Either this way, or early abort if we already reached the end.

Another topic, on the `m_` prefix. I actually thought about slightly moving towards how we name things in Contour, that is, using `_` preix for private member naming and continue in camelCase then (which also "slightly" matches what we use in the company). What do you think?

src/tests/DataMapperTests.cpp
```diff
@@ -111,6 +111,32 @@ TEST_CASE_METHOD(SqlTestFixture, "partial row retrieval", "[DataMapper]")
...
+        auto person = Person {};
+        person.name = "John";
+        person.age = i;
+        dm.Create(person);
+    }
+
+    auto stmt = SqlStatement { dm.Connection() };
+    int age = 40;
+    int id = 1;
+    for (auto&& person: iterator<Person>(stmt, "Person"))
```
christianparpart, line 130: since Lightweight library currently lives in the global namespace, I think, the type names should ideally be a bit more specific. `iterator` is a little bit too generic. IIUIC, what about `SqlRowIterator<Record>`?

src/Lightweight/SqlStatement.hpp
Yaraslaut: I agree, initially i used `_` as a prefix and then changed it to `m_` xD, will change it back

PR #109, opened 2024-11-27
SqlConnection: Default to SQL connection string and be more explicit about it

christianparpart:
FYI: we had serious issues with that gDefaultConnectionInfo-global when this library was built as DLL on Windows. So we decided to be more explicit about it, only allow connection strings as default way, but still allow data sources to be passed (explicitly only) and converted to connection strings, when explicitely requested.

### DIFF
--- a/src/Lightweight/SqlConnection.cpp
+++ b/src/Lightweight/SqlConnection.cpp
@@ -9,7 +9,7 @@
 using namespace std::chrono_literals;
 using namespace std::string_view_literals;
 
-static SqlConnectInfo gDefaultConnectInfo {};
+static SqlConnectionString gDefaultConnectionString {};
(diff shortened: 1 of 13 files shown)


src/Lightweight/SqlConnection.cpp
```diff
@@ -118,88 +127,53 @@ void SqlConnection::ResetPostConnectedHook()
...
-        PostConnect();
-
-        SqlLogger::GetLogger().OnConnectionOpened(*this);
-
-        if (gPostConnectedHook)
-            gPostConnectedHook(*this);
-
-        return true;
-    }
+    m_data->connectInfo = std::move(sqlConnectionString);
```
Yaraslaut, line 176: maybe this member also should become `conenctString` instead of `connectInfo` ?

PR #124, opened 2024-12-06
Reduce CREATE TABLE code duplication by making use of the query builder in data mapper's CreateTable<>

### LINKED ISSUE #121: DataMapper::CreateTable<> should use migration builder API to reduce code duplication
we currently have two was to `CREATE TABLE`'s.

Make use of the migration query builder API to create the tables using `DataMapper::CreateTable<Reccord>()` and related.

christianparpart:
reducing code duplication

closes #121.

### DIFF
--- a/src/Lightweight/SqlQuery/MigrationPlan.hpp
+++ b/src/Lightweight/SqlQuery/MigrationPlan.hpp
@@ -3,6 +3,8 @@
 #pragma once
 
 #include "../Api.hpp"
+#include "../SqlDataBinder.hpp"
+#include "../Utils.hpp"
 
 #include <reflection-cpp/reflection.hpp>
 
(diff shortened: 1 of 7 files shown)


src/Lightweight/SqlQuery/MigrationPlan.hpp
```diff
@@ -52,6 +53,134 @@ using SqlColumnTypeDefinition = std::variant<
 >;
 // clang-format on
 
+namespace detail
+{
+
+template <typename T>
+constexpr bool AlwaysFalse = false;
```
Yaraslaut, line 60: I prefer to use std::false_type for compile time context

src/Lightweight/SqlQuery/MigrationPlan.hpp
```diff
@@ -52,6 +53,134 @@ using SqlColumnTypeDefinition = std::variant<
...
+
+template <typename>
+struct SqlColumnTypeDefinitionOf;
```
Yaraslaut, line 63: if add `static_assert(false)` inside default instantiation then SFINAE error will be much better

src/Lightweight/SqlQuery/MigrationPlan.hpp
```diff
@@ -52,6 +53,134 @@ using SqlColumnTypeDefinition = std::variant<
...
+struct SqlColumnTypeDefinitionOf<int16_t>
+{
+    static constexpr auto value = SqlColumnTypeDefinitions::Smallint {};
+};
+
+template <>
+struct SqlColumnTypeDefinitionOf<uint16_t>
+{
+    static constexpr auto value = SqlColumnTypeDefinitions::Smallint {};
+};
```
Yaraslaut, line 117: Instead of two separate entries better to create specialization like:
```suggestion

template <typename T>
requires one_of<T, uint16_t, int16_t>
struct SqlColumnTypeDefinitionOf<T>
{
    static constexpr auto value = SqlColumnTypeDefinitions::Smallint {};
};
```

Yaraslaut (commented):
Some small comments

src/Lightweight/SqlQuery/MigrationPlan.hpp
christianparpart: true, but I think i still need it templated.

src/Lightweight/SqlQuery/MigrationPlan.hpp
christianparpart: i did that for the data binder. the problem I then faced is, that I did not see anymore what type was not implemented (i.e. which type did not have a template specialization for ). You will receive a compile error due to the static assert, which was great, but it did not tell me what type. The AlwaysFalse<T> should work around it (now that I write it out here :D). I'll try that.

src/Lightweight/SqlQuery/MigrationPlan.hpp
christianparpart: using `same_as... || ...` for now, because OneOf<> needs to be moved first to a more central helper file (tbd later).

src/Lightweight/SqlQuery/MigrationPlan.hpp
christianparpart: done

src/Lightweight/SqlQuery/MigrationPlan.hpp
christianparpart: done (via `AlwaysFalse<T>`)

src/Lightweight/SqlQuery/MigrationPlan.hpp
christianparpart: Done. Moved `OneOf<T, U...>` to Utils.hpp and rewrote the bits here.

Yaraslaut (approved):
Thanks :)

PR #125, opened 2024-12-07
Add input binder support & test cases for string_view/wstring_view/u16string_view/u32string_view

christianparpart:
tests tests tests

ok, i've changed more than initially anticipated (to just get the string view variants working), but I realized that we did not actually create tables with NCHAR and NVARCHAR data types, which is now fixed.

Oracle SQL got its own dialect formatter (not fully correct one), but this should be fixed by the time we add it to the CI.


FelixTheC (approved):
LGTM

Yaraslaut (approved):
LGTM

PR #126, opened 2024-12-07
Drop SqlTrimmedString in favor of SqlTrimmedFixedString<N>

christianparpart:
`SqlTrimmedString` is not necessarily needed, because `SqlTrimmedFixedString<N>` is the same thing plus it encodes its maximum length, does not need memory allocations beyond the object itself, and can be trivially pre-bound as output parameter.


Yaraslaut (approved):
Great :)

PR #129, opened 2024-12-09
Update ddl2cpp and add tests

Yaraslaut:
Adapt ddl2cpp to use new DataMapper and add tests


src/benchmark/tables.hpp
```diff
@@ -0,0 +1,89 @@
+#include <Lightweight/DataMapper/DataMapper.hpp>
```
christianparpart, line 1: `#pragma once` missing, for it being a header file

could we also somehow make it add a SPDX license tag, defaulting to Apache-2.0, but letting ddl2cpp's CLI arguments parametrize it?

src/benchmark/tables.hpp
```diff
@@ -0,0 +1,89 @@
...
+#include <Lightweight/SqlConnection.hpp>
+#include <Lightweight/SqlDataBinder.hpp>
+#include <Lightweight/SqlQuery.hpp>
+#include <Lightweight/SqlQueryFormatter.hpp>
+#include <Lightweight/SqlScopedTraceLogger.hpp>
+#include <Lightweight/SqlStatement.hpp>
+#include <Lightweight/SqlTransaction.hpp>
+struct lists;
```
christianparpart, line 9:
```suggestion

struct lists;
```

just a readability thought

src/tools/ddl2cpp.cpp
```diff
@@ -47,7 +48,7 @@ std::string MakeType(SqlSchema::Column const& column)
             else
                 return std::format("std::string", column.size);
         case ColumnType::TEXT:
-            return std::format("SqlText");
+            return std::format("SqlString<30>");
         case ColumnType::BOOLEAN:
```
christianparpart, line 52: LNB: Due to the lag of information detailness of `enum class ColumnType`, it'll soonish be replaced by `SqlColumnTypeDeclaration` (`variant` over simple structs), which makes it easy to encode text length along with it.

src/tools/ddl2cpp.cpp
```diff
@@ -160,33 +170,27 @@ class CxxModelPrinter
...
+            }
             if (column.isForeignKey)
                 continue;
-            m_definitions << std::format("    Model::Field<{}, {}, \"{}\"{}> {};\n",
-                                         type,
-                                         columnPosition,
-                                         column.name,
-                                         column.isNullable ? ", Nullable" : "",
-                                         column.name);
+            m_definitions << std::format("    Field<{}> {};\n", type, column.name);
```
christianparpart, line 179: nullable fields are `std::optional<T>` rather than `T` alone. Is this covered?

src/tools/tests/1.sq
```diff
@@ -0,0 +1,6 @@
+CREATE TABLE "Person" (
```
christianparpart, line 1: why are those file names ending with `.sq` and not `.sql`?

Yaraslaut: Changed

src/tools/ddl2cpp.cpp
Yaraslaut: Fixed

src/benchmark/tables.hpp
Yaraslaut: I added now Apache-2.0 header, can extend it in the future PRs

src/tools/ddl2cpp.cpp
Yaraslaut: Returned SqlText for now

christianparpart (approved):
Thank you very much

PR #130, opened 2024-12-09
feedback oriented improvements

christianparpart:
# most important additions:

* ddl2cpp: Drop dead `MakePluralVariableName(table)`
* Adapt `BasicStringBinder<T>::CharType` to work with non-stdlib like strings
* QueryBuilder: Extend `Where(A[, OP], B)` to also allow `B` to be a `SqlQualifiedTableColumnName`
* QueryBuilder: Add intuitive way to test for `NULL`: `Where(A, SqlNullValue)` and `Where(A, std::nullopt)`

### DIFF
--- a/src/tests/QueryBuilderTests.cpp
+++ b/src/tests/QueryBuilderTests.cpp
@@ -456,8 +456,8 @@ TEST_CASE_METHOD(SqlTestFixture, "Where: left IS NULL", "[SqlQueryBuilder]")
             return q.FromTable("That")
                 .Select()
                 .Field("foo")
-                .Not().Where("Left1", SqlNullValue)
-                .Or().Not().Where("Left2", std::nullopt)
+                .WhereNotEqual("Left1", SqlNullValue)
+                .Or().WhereNotEqual("Left2", std::nullopt)
                 .All();
             // clang-format on
         },
--- a/src/tests/QueryBuilderTests.cpp
+++ b/src/tests/QueryBuilderTests.cpp
@@ -418,6 +418,24 @@ TEST_CASE_METHOD(SqlTestFixture, "SqlQueryBuilder.WhereColumn", "[SqlQueryBuilde
                                   WHERE "left" = "right")"));
 }
 
+TEST_CASE_METHOD(SqlTestFixture,
+                 "Where: SqlQualifiedTableColumnName OP SqlQualifiedTableColumnName",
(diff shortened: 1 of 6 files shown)


src/tests/QueryBuilderTests.cpp
```diff
@@ -418,6 +418,53 @@ TEST_CASE_METHOD(SqlTestFixture, "SqlQueryBuilder.WhereColumn", "[SqlQueryBuilde
...
+        },
+        QueryExpectations::All(R"(SELECT "foo" FROM "That"
+                                  WHERE "Left1" IS NULL AND "Left2" IS NULL)"));
+
+    checkSqlQueryBuilder(
+        [](SqlQueryBuilder& q) {
+            return q.FromTable("That")
+                .Select()
+                .Field("foo")
+                .Not().Where("Left1", SqlNullValue)
```
Yaraslaut, line 460: I think this is not the most intuitive API, something like 
```suggestion
                .Where("Left1").Not(SqlNullValue)
```
is better, and can distinguish between complete/incomplete builder state

Yaraslaut (changes requested):
I think it is better to change API slightly

src/tests/QueryBuilderTests.cpp
christianparpart: I agree. There is also `WhereNot(value)` though. We could add a `WhereNotEquals(a, b)` express inequality.

`Where(a).Not(b)` I fear, might be hard to read too. We should focus on more expressive function member names instead. We already have `WhereNull` and `WhereNotNull` for example.

Yaraslaut: so we can hide `Where(a).Not(b)` as implementation detail and provide `WhereNotEquals(a,b)` for the user, what do you think?

christianparpart: @Yaraslaut I've been adding `WhereNotEqual(a, b)` and adapted the tests accordingly. There is no need to add `Where(a).Not(b)` as an implementation detail. (I also cannot think of a way to implement Where(a).Not(b) without breaking current Where(a) semantics).

P.s.: I was generally getting inspired on the query builder API by the Dapper project: https://github.com/DapperLib/Dapper :)

Yaraslaut (approved):
Thanks

Issue #116, opened 2024-12-02
extend migration API for foreign key support and alter column type / null / not null


christianparpart:
- [x] CreateTable: foreign keys
- [x] AlterTable: foreign keys
- [x] AlterTable: ChangeColumn type
- [x] AlterTable: ChangeColumn null / not null

refs #108

PR #133, opened 2024-12-10
Add foreign key handling in  SqlQuery

Yaraslaut:
Refs https://github.com/LASTRADA-Software/Lightweight/issues/116 
Implements CreateTable: foreign keys

### DIFF
--- a/src/Lightweight/SqlQuery/MigrationPlan.hpp
+++ b/src/Lightweight/SqlQuery/MigrationPlan.hpp
@@ -155,11 +155,18 @@ enum class SqlPrimaryKeyType : uint8_t
     GUID,
 };
 
+struct SqlForeignKeyReferenceDefinition
+{
+    std::string tableName;
+    std::string columnName;
+};
+
 struct SqlColumnDeclaration
 {
     std::string name;
     SqlColumnTypeDefinition type;
     SqlPrimaryKeyType primaryKey { SqlPrimaryKeyType::NONE };
+    std::optional<SqlForeignKeyReferenceDefinition> foreignKey {};
     bool required { false };
     bool unique { false };
     bool index { false };
--- a/src/Lightweight/SqlQuery/Migrate.cpp
+++ b/src/Lightweight/SqlQuery/Migrate.cpp
@@ -147,6 +147,18 @@ SqlCreateTableQueryBuilder& SqlCreateTableQueryBuilder::PrimaryKeyWithAutoIncrem
     });
 }
 
+SqlCreateTableQueryBuilder& SqlCreateTableQueryBuilder::ForeignKey(std::string columnName,
+                                                                   SqlColumnTypeDefinition columnType,
+                                                                   SqlForeignKeyReferenceDefinition foreignKey)
+{
+    return Column(SqlColumnDeclaration {
+        .name = std::move(columnName),
+        .type = columnType,
+        .foreignKey = std::move(foreignKey),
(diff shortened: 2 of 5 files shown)


src/Lightweight/DataMapper/BelongsTo.hpp
```diff
@@ -217,3 +217,11 @@ struct SqlDataBinder<BelongsTo<ReferencedField>>
         return sqlReturn;
     }
 };
+
+// TODO Something is wrong with the ValueType of BelongsTo. It should be the same as the primary key of the other
+// record. Which is Integer, but SqlColumnTypeDefinitionOf deduce it as Bigint.
+template <auto F>
+struct detail::SqlColumnTypeDefinitionOf<BelongsTo<F>>
+{
+    static constexpr auto value = SqlColumnTypeDefinitions::Integer {};
```
christianparpart, line 226: eh, I think the value's type should be inferred from `BelongsTo<F>` itself. When that points to `&Other::field`, and that field is e.g. a `Guid`, then we should map to `SqlColumnTypeDefinitions::Guid {}` here. I think i tmight end up being possible with the inline lambda function call trick `... = []() constexpr -> SqlColumnTypeDefinition { if constexpr (x) return x'; ... else return y'; }();`

src/Lightweight/SqlQuery/MigrationPlan.hpp
```diff
@@ -155,11 +155,18 @@ enum class SqlPrimaryKeyType : uint8_t
     GUID,
 };
 
+struct SqlForeignKeyDefinition
```
christianparpart, line 158:
```suggestion
struct SqlForeignKeyReferenceDefinition
```
I think a proper term should include that we're referring to the reference part of a foreign key here. it's the other end of an FK only :)

src/Lightweight/DataMapper/BelongsTo.hpp
Yaraslaut: This is usually is possible with recursive consteval lambdas, issue here is that i was getting wrong type from BelongsTo<>::ValueType itself, Bigint instead of integer for primary key of another structure, we also can restrict instantiotion of BelongsTo only for PrimaryKeys, and then we know type of the field

christianparpart: @Yaraslaut I think `Bigint` should be actually more correct than `Integer` (even though I think both should be supported), but integer PKs can easily explode beyond integer's max value, and hence, bigint should be used anyways.

On the other hand (see other PR), we should be prepared to expect GUID column types as PK as well. At least for those PKs that are subject to be referenced by foreign keys.

christianparpart:
@Yaraslaut can we split this PR into two? One for the query builder and one for ddl2cpp's use?

Because the query builder api changes can definitely be merged early.

Yaraslaut:
Sure, I will split them

src/Lightweight/DataMapper/BelongsTo.hpp
Yaraslaut: I will adapt on top of guid merge, and we can then use bigint that is automatically deduced

christianparpart: i've merged the PR, because I assume there was/is no veto in it. And nobody is held back. :)

christianparpart (approved):
Many thanks

Issue #132, opened 2024-12-10
OR condition in a concated join statement


FelixTheC:
I would be good to be able to create Join's with an OR clause.

```cpp
      auto query = queryStmt.FromTable("BETON_SORTE")
                            .Select()
                            .LeftOuterJoin("BEN_DESIGN"sv, "SORTE_NR"sv, "NR"sv)
                            .RightOuterJoin("BETON_ZEMENTART"sv, [](SqlJoinConditionBuilder q)
                            {
                                return q.On("NR", { .tableName = "BETON_SORTE", .columnName = "ZEMENTART_NR" })
                                        .Or("NR", { .tableName = "BETON_SORTE", .columnName = "ZEM_2_NR" })
                                        .Or("NR", { .tableName = "BETON_SORTE", .columnName = "ZEM_3_NR" })
                                        .Or("NR", { .tableName = "BETON_SORTE", .columnName = "ZEM_4_NR" });
                            })
                            .Fields({ "NR"sv, "NAME"sv }, "BETON_ZEMENTART");
```

the SQL Statement should then look like this

```sql
SELECT "TableA"."NR", "TableA"."NAME" FROM "TableA"
    LEFT OUTER JOIN "TableB" ON "TableA"."NR" = "TableB"."ZEMENTART_NR" 
        OR "TableA"."NR" = "TableB"."ZEM_2_NR" 
        OR "TableA"."NR" = "TableB"."ZEM_3_NR" 
        OR "TableA"."NR" = "TableB"."ZEM_4_NR"
    LEFT OUTER JOIN "TableC" ON "TableC"."SORTE_NR" = "TableB"."NR"
```

PR #134, opened 2024-12-10
Add OrOn operator for SqlJoinConditionBuilder

Yaraslaut:
Closes https://github.com/LASTRADA-Software/Lightweight/issues/132

### DIFF
--- a/src/tests/QueryBuilderTests.cpp
+++ b/src/tests/QueryBuilderTests.cpp
@@ -216,6 +216,31 @@ TEST_CASE_METHOD(SqlTestFixture, "SqlQueryBuilder.FieldsWithBelongsTo", "[SqlQue
         });
 }
 
+TEST_CASE_METHOD(SqlTestFixture, "SqlQueryBuilder.ComplexOR", "[SqlQueryBuilder]")
+{
+    checkSqlQueryBuilder(
+        [](SqlQueryBuilder& q) {
+            using namespace std::string_view_literals;
+            return q.FromTable("Table1")
+                .Select()
+                .LeftOuterJoin("Table2"sv, "id"sv, "id"sv)
+                .RightOuterJoin("Table3"sv,
+                                [](SqlJoinConditionBuilder q) {
+                                    // clang-format off
+                                    return q.On("id", { .tableName = "Table1", .columnName = "column1" })
+                                            .OrOn("id", { .tableName = "Table1", .columnName = "column2" })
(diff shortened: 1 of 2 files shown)


FelixTheC:
@Yaraslaut I had a similar Idea but @christianparpart thought about creating a parent class for the Where and Join Statements as they start sharing the same behavior. Like `And`, `Or` maybe we should also include `Not`.

src/tests/QueryBuilderTests.cpp
```diff
@@ -216,6 +216,29 @@ TEST_CASE_METHOD(SqlTestFixture, "SqlQueryBuilder.FieldsWithBelongsTo", "[SqlQue
...
+            using namespace std::string_view_literals;
+            return q.FromTable("BETON_SORTE")
+                .Select()
+                .LeftOuterJoin("BEN_DESIGN"sv, "SORTE_NR"sv, "NR"sv)
+                .RightOuterJoin("BETON_ZEMENTART"sv,
+                                [](SqlJoinConditionBuilder q) {
+                                    return q.On("NR", { .tableName = "BETON_SORTE", .columnName = "ZEMENTART_NR" })
+                                        .Or("NR", { .tableName = "BETON_SORTE", .columnName = "ZEM_2_NR" })
+                                        .Or("NR", { .tableName = "BETON_SORTE", .columnName = "ZEM_3_NR" })
+                                        .Or("NR", { .tableName = "BETON_SORTE", .columnName = "ZEM_4_NR" });
```
christianparpart, line 232:
```suggestion
                                        .OrOn("NR", { .tableName = "BETON_SORTE", .columnName = "ZEM_2_NR" })
                                        .OrOn("NR", { .tableName = "BETON_SORTE", .columnName = "ZEM_3_NR" })
                                        .OrOn("NR", { .tableName = "BETON_SORTE", .columnName = "ZEM_4_NR" });
```

Analogous to `Where` vs `OrWhere` - what do you two guys think about `On` vs `OrOn`?

Sounds more homogenous to me than `On` vs `Or`.

Yaraslaut:
Ah, I see, it just this is a small changes, some repetition exist in query builder but it can be adapted later when api is more stable

src/tests/QueryBuilderTests.cpp
Yaraslaut: will change

christianparpart (approved):
I've rewritten the test case to not contain any business schema but rather generic table / column names.

PR #137, opened 2024-12-10
Add SqlMigration API

christianparpart:
this is prep work of providing a CLI database migration tool, similar to what other well known SQL frameworks provide as well (me being inspired here by e.g. Rails way of migrating)


christianparpart:
merging early due to conflicts in upcoming docs PR. remaining migration diffs are coming later then.

PR #139, opened 2024-12-13
Add BelongsTo handling in `DataMapper::CreateTableString<T>()`

Yaraslaut:


christianparpart:
```
  42000 (3726) - [Microsoft][ODBC Driver 18 for SQL Server][SQL Server]Could
  not drop object 'User' because it is referenced by a FOREIGN KEY constraint.
```

I think here are two ways to fix this when dropping all tables in the CI fixture:

1. We either use Fixed-Point Iteration style algorithm until all tables are dropped, e.g. re-iterate over all not-yet-dropped tables until they are dropped. this way you'll impliictely have deleded all referencing tables first. 
2. Or we actually inspect the tables to be dropped, and recursively drop the referencing tables first, like a `DropTableRecursively<Record>()`. AFAICT, we should be even able to solve the order of deletion at compile time, because we have all the info at compile time already. Similar to how we find the primary keys / foreign keys in a table struct right now already, too.

Having that one written out, I think the second solution is certainly way more sexy, but involves more thinking.

Yaraslaut:
I will go with second option, will update PR soon

src/Lightweight/DataMapper/DataMapper.hpp
```diff
@@ -315,12 +324,20 @@ std::vector<std::string> DataMapper::CreateTableString(SqlServerType serverType)
...
+
+    // Add Foreing keys from BelongsTo entries
+    Reflection::EnumerateMembers<Record>([&]<size_t I, typename FieldType>() {
+        if constexpr (IsBelongsTo<FieldType>)
+        {
+            createTable.ForeignKey(
+                std::string(FieldNameOf<I, Record>()),
+                SqlColumnTypeDefinitionOf<typename FieldType::ValueType>,
+                SqlForeignKeyReferenceDefinition { .tableName = std::string{RecordTableName<typename FieldType::ReferencedRecord>},
+                                                   .columnName = "id" });
```
christianparpart, line 340: I think `"id"` may potentially be wrong. Because the foreign key's referenced column name is not mandated to be named `"id"`. :thinking:

src/tests/DataMapperTests.cpp
```diff
@@ -243,6 +243,23 @@ TEST_CASE_METHOD(SqlTestFixture, "BelongsTo", "[DataMapper][relations]")
...
+
+    if (dm.Connection().ServerType() == SqlServerType::SQLITE)
+    {
+
+        REQUIRE(dm.CreateTableString<User>(dm.Connection().ServerType())[0]
+                == "CREATE TABLE \"User\" (\n"
+                   "    \"id\" INTEGER NOT NULL PRIMARY KEY AUTOINCREMENT,\n"
+                   "    \"name\" VARCHAR(30) NOT NULL\n"
+                   ");");
+        REQUIRE(dm.CreateTableString<Email>(dm.Connection().ServerType())[0]
```
christianparpart, line 255: instead of calling `[0]` here, I think we should instead put both `NormalizeText(..)` helper methods into tests/Utils.hpp and use those instead.

christianparpart (changes requested):

src/Lightweight/DataMapper/DataMapper.hpp
Yaraslaut: Fixed, now we check for primary key name

src/tests/DataMapperTests.cpp
Yaraslaut: Switched to `NormalizeText` function

src/Lightweight/DataMapper/DataMapper.hpp
```diff
@@ -345,6 +347,33 @@ std::vector<std::string> DataMapper::CreateTableString(SqlServerType serverType)
...
+                    .tableName = std::string { RecordTableName<typename FieldType::ReferencedRecord> },
+                    .columnName = std::string { FieldNameOf<[]() consteval {
+                        size_t index = -1;
+                        Reflection::EnumerateMembers<typename FieldType::ReferencedRecord>(
+                            [&]<size_t J, typename ReferencedFieldType>() consteval {
+                                if constexpr (IsField<ReferencedFieldType>)
+                                {
+                                    if constexpr (ReferencedFieldType::IsPrimaryKey)
+                                    {
+                                        index = J;
```
christianparpart, line 367:
```suggestion
                                        return J;
```

src/Lightweight/DataMapper/DataMapper.hpp
```diff
@@ -345,6 +347,33 @@ std::vector<std::string> DataMapper::CreateTableString(SqlServerType serverType)
...
+    Reflection::EnumerateMembers<Record>([&]<size_t I, typename FieldType>() {
+        if constexpr (IsBelongsTo<FieldType>)
+        {
+            createTable.ForeignKey(
+                std::string(FieldNameOf<I, Record>),
+                SqlColumnTypeDefinitionOf<typename FieldType::ValueType>,
+                SqlForeignKeyReferenceDefinition {
+                    .tableName = std::string { RecordTableName<typename FieldType::ReferencedRecord> },
+                    .columnName = std::string { FieldNameOf<[]() consteval {
+                        size_t index = -1;
```
christianparpart, line 360:
```suggestion
```

src/Lightweight/DataMapper/DataMapper.hpp
```diff
@@ -345,6 +347,33 @@ std::vector<std::string> DataMapper::CreateTableString(SqlServerType serverType)
...
+                            [&]<size_t J, typename ReferencedFieldType>() consteval {
+                                if constexpr (IsField<ReferencedFieldType>)
+                                {
+                                    if constexpr (ReferencedFieldType::IsPrimaryKey)
+                                    {
+                                        index = J;
+                                    }
+                                }
+                            });
+                        return index;
```
christianparpart, line 371: and then, return here something symbolic to signal that nothing was found.
Giving it a second thought. I think this should never be possible (or is it?) In that case, we could `static_assert(false, "reason");` here, assuming that this works on all 3 compilers, latest version.

src/tests/DataMapperTests.cpp
```diff
@@ -268,6 +268,23 @@ TEST_CASE_METHOD(SqlTestFixture, "BelongsTo", "[DataMapper][relations]")
 
     actualEmail1.user.Unload();
     REQUIRE(!actualEmail1.user.IsLoaded());
+
+    if (dm.Connection().ServerType() == SqlServerType::SQLITE)
```
christianparpart, line 272: Why only if Sqlite? Or is it just to see if it generally works?

src/tests/DataMapperTests.cpp
```diff
@@ -500,6 +517,35 @@ TEST_CASE_METHOD(SqlTestFixture, "HasManyThrough", "[DataMapper][relations]")
     CHECK(patient.physicians.Count() == 2);
     CHECK(patient.physicians.At(0).name.Value() == "Granny");
     CHECK(DataMapper::Inspect(patient.physicians.At(0)) == DataMapper::Inspect(physician2));
+
+    if (dm.Connection().ServerType() == SqlServerType::SQLITE)
```
christianparpart, line 521: squared :)

src/Lightweight/DataMapper/DataMapper.hpp
Yaraslaut: I think this is not possible since you can not return out of scope of `Reflection::EnumerateMembers` the only way to return something from there is through lambda capture

src/tests/DataMapperTests.cpp
Yaraslaut: just to see that everything is working, did not want to bother writing comparison for every database type

src/Lightweight/DataMapper/DataMapper.hpp
christianparpart: brain damage on my end. apologees. need to rethink :)

src/Lightweight/DataMapper/DataMapper.hpp
Yaraslaut: the default value of -1 ensures that if we never find primary key compilation will fail since struct member will not exist

src/Lightweight/DataMapper/DataMapper.hpp
```diff
@@ -283,12 +283,14 @@ template <std::size_t I, typename Record>
...
-    static constexpr auto storage = []() -> std::array<char, baseName.size() + 3> {
+    static constexpr auto storage = []() -> std::array<char, baseName.size() + 3>
+    {
         std::array<char, baseName.size() + 3> storage;
         std::copy_n(baseName.begin(), baseName.size(), storage.begin());
         std::copy_n("_id", 3, storage.begin() + baseName.size());
         return storage;
-    }();
+    }
+    ();
```
christianparpart, line 293: is this really the syntax of clang-format? :thinking:

src/tests/DataMapperTests.cpp
```diff
@@ -268,6 +268,23 @@ TEST_CASE_METHOD(SqlTestFixture, "BelongsTo", "[DataMapper][relations]")
...
+    {
+
+        REQUIRE(NormalizeText(dm.CreateTableString<User>(dm.Connection().ServerType()))
+                == "CREATE TABLE \"User\" (\n"
+                   "\"id\" INTEGER NOT NULL PRIMARY KEY AUTOINCREMENT,\n"
+                   "\"name\" VARCHAR(30) NOT NULL\n"
+                   ");");
+        REQUIRE(NormalizeText(dm.CreateTableString<Email>(dm.Connection().ServerType()))
+                == "CREATE TABLE \"Email\" (\n"
```
christianparpart, line 281: I think it's better to use raw string literals here rather than many string literals that need escaping. Should be easier to read, isn't it?

NormalizeText() takes care of not having to deal with indentation width etc.

Yaraslaut: Fixed

src/Lightweight/DataMapper/DataMapper.hpp
```diff
@@ -336,6 +338,26 @@ std::vector<std::string> DataMapper::CreateTableString(SqlServerType serverType)
             else if constexpr (FieldType::IsPrimaryKey)
                 createTable.PrimaryKey(std::string(FieldNameOf<I, Record>),
                                        SqlColumnTypeDefinitionOf<typename FieldType::ValueType>);
+            else if constexpr (IsBelongsTo<FieldType>)
+            {
+                constexpr size_t referencedFieldIndex = []() constexpr -> size_t {
+                    size_t ind = -1;
```
christianparpart, line 344: One last comment, just for the record, and similar to how we should dothings in Contour (of course :D). Let's try to avoid abbreviations unless they're very well known. :)

christianparpart (approved):
Many thanks :christmas_tree:

src/Lightweight/DataMapper/DataMapper.hpp
Yaraslaut: renamed it into `index`

PR #142, opened 2024-12-14
[DataMapper] Field: Find better names for the PrimaryKey enum class members

christianparpart:
this is in reference to a feedback from @FelixTheC, to find better names for the enum class emmbers, specifically the `Manual` name, that wasn't actually that manual, but more like semi automatic.

I hope this one are now better fitting names.


Yaraslaut (approved):
LGTM

PR #145, opened 2024-12-14
[DataMapper] Create: Small refactor to reduce the visual load of calling `IsModified()`

christianparpart:
Also add an IsAutoAssignPrimaryKey helper to Field.

Based on feedback by @StepKie.

### DIFF
--- a/src/Lightweight/DataMapper/DataMapper.hpp
+++ b/src/Lightweight/DataMapper/DataMapper.hpp
@@ -402,17 +402,16 @@ RecordId DataMapper::Create(Record& record)
 
     // If the primary key is not an auto-increment field and the primary key is not set, we need to set it.
     CallOnPrimaryKey(record, [&]<size_t PrimaryKeyIndex, typename PrimaryKeyType>(PrimaryKeyType& primaryKeyField) {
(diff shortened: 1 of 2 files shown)


src/Lightweight/DataMapper/DataMapper.hpp
```diff
@@ -402,17 +402,16 @@ RecordId DataMapper::Create(Record& record)
 
     // If the primary key is not an auto-increment field and the primary key is not set, we need to set it.
     CallOnPrimaryKey(record, [&]<size_t PrimaryKeyIndex, typename PrimaryKeyType>(PrimaryKeyType& primaryKeyField) {
-        if constexpr (!PrimaryKeyType::IsAutoIncrementPrimaryKey)
+        if (!primaryKeyField.IsModified())
```
Yaraslaut, line 405: I am not sure that this change of conditions gives any benefits, maybe it is better to have `if constexpr` in the most outer scope ?

christianparpart: check update. like so? :)

Yaraslaut (approved):
Nice, thanks :)

Issue #136, opened 2024-12-10
`Field<T>` should optionally allow passing an SQL column name, like `Field<T, Name>`


christianparpart:
We currently infer the SQL column name by compile-time reflecting the variable name of the given field.

This is good, but sometimes on already existing databases, the table and column names are not nicely chosen and that would make it hard to read on the C++ side (or inconvenient).

The idea is to allow passing an SQL name override as template parameter, e.g. like so:

```cpp
struct Person
{
    Field<uint64_t, "NR"> id;          // SQL column name is "NR"
    Field<SqlString<30>, "NAME"> name; // SQL column name is "NAME"
    Field<SqlString<40>> email;         // SQL column name is "email", as inferred by the member name
};
```

christianparpart:
@FelixTheC does this make sense to you?

FelixTheC:
@christianparpart looks good I really had something like this in my mind
https://docs.pydantic.dev/latest/concepts/alias/

christianparpart:
> @christianparpart looks good I really had something like this in my mind https://docs.pydantic.dev/latest/concepts/alias/

It should actually be also possible like so:

```cpp
Field<int, SqlColumnAlias<"BLAH">> field;
```

PR #146, opened 2024-12-14
cleanups plus SQL column name aliasing

christianparpart:
cleanups, but also closing #136.

### DIFF
--- a/src/tests/DataMapperTests.cpp
+++ b/src/tests/DataMapperTests.cpp
@@ -20,8 +20,8 @@ std::ostream& operator<<(std::ostream& os, RecordId const& id)
     return os << id.value;
 }
 
-template <typename T, PrimaryKey IsPrimaryKeyValue>
-std::ostream& operator<<(std::ostream& os, Field<std::optional<T>, IsPrimaryKeyValue> const& field)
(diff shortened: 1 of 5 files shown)


src/tests/DataMapperTests.cpp
```diff
@@ -595,3 +595,38 @@ TEST_CASE_METHOD(SqlTestFixture, "Table with multiple primary keys", "[DataMappe
...
     CHECK(queriedRecord == record);
 }
+
+struct AliasedRecord
+{
+    Field<uint64_t, PrimaryKey::ServerSideAutoIncrement, SqlRealName { "pk" }> id {};
+    Field<SqlAnsiString<30>, SqlRealName { "c1" }> name;
+    Field<SqlAnsiString<30>, SqlRealName { "c2" }> comment;
+
+    static constexpr std::string_view TableName = "TheAliasedRecord";
```
christianparpart, line 605: @FelixTheC, You could already alias the SQL tabl ename. Now you can also alias the SQL column names as shown above.

Yaraslaut (approved):
the only thing worth changing is nullopt into monostate but this can be omitted since a small nit pick anyway, so nice addition

PR #147, opened 2024-12-15
Drop tables via fixed-point iteration

christianparpart:
... to naively ensure all tables are dropped, regardless of foreign key constraint priorities.

@Yaraslaut you might want that in #139 early, to get tables properly cleaned up in test suite.


Yaraslaut (approved):
Thanks

christianparpart:
@Yaraslaut i've pushed a pico change to FixedPOintIterate, to be slightly more efficient (less copying).

PR #153, opened 2024-12-17
some convenience helper for very complex queries

christianparpart:
We have a case where we add many many columns to the query builder and thus, chain each fields into `BindOutputColumn...`.

With this API, it shall be way more easy, concise, and less error prone, to construct and maintain such queries.

NB: check the test case for a usage example


christianparpart:
Oops. Auto merge did not wait for approval ;)

PR #155, opened 2024-12-19
Allow adding and dropping foreign keys via SQL Query Builder

christianparpart:
Refs #116


christianparpart:
i guess i need to change some settings. this wasn't expected to be merged without approval :D

PR #165, opened 2024-12-28
Add documentation for sql query builder

Yaraslaut:
Add documentation for sql query builder to cover some common use cases and have one place with a short list of available functions

### DIFF
--- a/src/Lightweight/SqlQuery/Migrate.hpp
+++ b/src/Lightweight/SqlQuery/Migrate.hpp
@@ -81,8 +81,8 @@ class [[nodiscard]] SqlAlterTableQueryBuilder final
     LIGHTWEIGHT_API SqlAlterTableQueryBuilder& AddColumn(std::string columnName, SqlColumnTypeDefinition columnType);
 
     /// Adds a new column to the table that is nullable.
-    LIGHTWEIGHT_API SqlAlterTableQueryBuilder& AddColumnAsNullable(std::string columnName,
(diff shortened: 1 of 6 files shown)


src/Lightweight/SqlQuery/Migrate.hpp
```diff
@@ -101,6 +103,7 @@ class [[nodiscard]] SqlAlterTableQueryBuilder final
     /// for (auto const& sql: sqlMigration)
     ///     stmt.ExecuteDirect(sql);
     /// @endcode
+    //TODO(PR) CHANGE NAME ?
```
christianparpart, line 106: What should we do about these TODO(PR) comments? Change name to what?

christianparpart (changes requested):
What do we do with the TODO(pr)'s?

src/Lightweight/SqlQuery/Migrate.hpp
Yaraslaut: We have different behavior for the CreateTable and AlterTable nested functions, for example while AddColumn in CreateTable makes it optional, AddColumn in AlterTable makes it required, so we need to have same behavior for this cases, that is why i left this TODO, to figure out where to rename to what

christianparpart: oh damn. remind me tomorrow, so we can talk about it in person. :thinking:

christianparpart (approved):
nice. thank you. :)

Issue #90, opened 2024-11-12
Allow comparing a record to another to collect the differences


christianparpart:
```cpp
template <typename Record>
DifferenceView<Record> CollectDifferences(Record const& left, Record const& right) noexcept;
```

where `DifferenceView<T>` holds a reference to the old and new value of each element that has changed from `left` to `right`.

### Example applications

This can be used to easily visualize a before/after state when changing a record in the table.

PR #166, opened 2025-01-01
Implement CollectDifferences

Yaraslaut:
Closes https://github.com/LASTRADA-Software/Lightweight/issues/90

Unfortunately we can not provide something like 
```
for(auto& [l,r] : diff)
...
```
since we can not deduce at runtime types :( 
so we need to use lambda that can iterate over fields with the differences


christianparpart (approved):
thank you

PR #170, opened 2025-01-07
minors found

christianparpart:


.github/workflows/build.yml
```diff
@@ -101,15 +101,23 @@ jobs:
...
         env:
           VCPKG_ROOT: "${{ runner.workspace }}/vcpkg"
       - name: "Build"
-        run: cmake --build --preset windows-cl-release
+        run: cmake --build --preset windows-cl-debug
+      - name: "DEBUG"
+        run: |
+          Get-ChildItem -Path out -Recurse -Filter '*.exe' | ForEach-Object { $_.FullName }
+          ./out/build/windows-cl-debug/target/Debug/LightweightTest.exe --help
+          ./out/build/windows-cl-debug/target/Debug/LightweightTest.exe -- --help
```
Yaraslaut, line 118: This takes to much time, better to run it after merge in master

Yaraslaut (commented):
LGTM, but we have to consider time for running actions, better to have them less than few minutes

.github/workflows/build.yml
christianparpart: I may need to disable it again, because I just can't get sqliteodbc automatically and correctly installed, apparently.

PR #177, opened 2025-01-15
DataMapper: Allow passing a classic (non-Field<>'ed members) struct to `QuerySingle<>`

christianparpart:


src/Lightweight/DataBinder/BasicStringBinder.hpp
```diff
@@ -114,6 +114,8 @@ struct LIGHTWEIGHT_API SqlDataBinder<AnsiStringType>
     {
         if constexpr (requires { AnsiStringType::Capacity; })
             StringTraits::Resize(result, AnsiStringType::Capacity);
+        else
```
christianparpart, line 117: To-do: Only resize to N if the buffer is currently 0.

Also for the Wide string equivalent.

Yaraslaut (approved):
LGTM, i added two cases to covert different number of elements in the struct and table

christianparpart:
> LGTM, i added two cases to covert different number of elements in the struct and table

oh nice, many thanks.

PR #178, opened 2025-01-16
Add support for std::u8string and std::u8string_view

### LINKED ISSUE #176: Add data binder for std::u8string and std::u8string_view
the data binder for `u8string` must of course also allow reading a column, while u8string_view can only be used for writing a column to the server.

extend the parametrized templated data binder to the two new string types.

christianparpart:
closes #176

### DIFF
--- a/src/Lightweight/DataBinder/BasicStringBinder.hpp
+++ b/src/Lightweight/DataBinder/BasicStringBinder.hpp
@@ -76,6 +76,59 @@ SQLRETURN GetColumnUtf16(SQLHSTMT stmt,
     return GetArrayData<SQL_C_WCHAR>(stmt, column, result, indicator);
 }
 
+template <typename StringType>
+SQLRETURN OutputColumnNonUtf16Unicode(
+    SQLHSTMT stmt, SQLUSMALLINT column, StringType* result, SQLLEN* indicator, SqlDataBinderCallback& cb) noexcept
+{
+    using CharType = typename StringType::value_type;
+
+    auto u16String = std::make_shared<std::u16string>();
+    if (!result->empty())
+        u16String->resize(result->size());
+    else
+        u16String->resize(255);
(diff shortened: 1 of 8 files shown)


src/Lightweight/DataBinder/BasicStringBinder.hpp
```diff
@@ -481,3 +481,125 @@ struct LIGHTWEIGHT_API SqlDataBinder<Utf32StringType>
...
+        });
+
+        return SQLBindCol(stmt,
+                          column,
+                          CType,
+                          static_cast<SQLPOINTER>(u16String->data()),
+                          static_cast<SQLLEN>(u16String->size() * sizeof(char16_t)),
+                          indicator);
+
+    }
```
Yaraslaut, line 584: I think we can make such function generic by providing string type and using ::char_type trait for the underlying char type, such that in this specialization we can use 
```
   static SQLRETURN OutputColumn(SQLHSTMT stmt,
                                  SQLUSMALLINT column,
                                  Utf8StringType* result,
                                  SQLLEN* indicator,
                                  SqlDataBinderCallback& cb) noexcept
  {
                 return OutputColumn<Utf8string, std::u16string>(stmpt, column, result, indicator, cb);     
  }

Yaraslaut (commented):
Just a small comment

src/Lightweight/DataBinder/BasicStringBinder.hpp
christianparpart: thanks, yeah i was thinking about that, too. also to reduce room of error and diverging implementations. :thinking:.

christianparpart: should be working now.

Yaraslaut (approved):
LGTM, thanks

Issue #180, opened 2025-01-18
Eliminate `SqlTraits` and `SqlColumnType`


christianparpart:
Will be wanted to be eliminated before we start on #172.

PR #181, opened 2025-01-18
Code cleanups

christianparpart:
This is mainly about dropping `SqlColumnType` and having its uses replaced with `SqlColumnTypeDefinition`

refs #180 
refs #182 

### TODO:

- [x] also, make sure to run ddl2cpp test suite on CI
- [ ] anything missing? ...


src/Lightweight/SqlSchema.cpp
```diff
@@ -251,6 +251,8 @@ void ReadAllTables(std::string_view database, std::string_view schema, EventHand
             // 12 - remarks
             column.defaultValue = columnStmt.GetColumn<std::string>(13);
 
+            column.type = FromNativeDataType(type, column.size, column.decimalDigits).value();
```
Yaraslaut, line 254: if we return optional, better to use here `value_or`

Yaraslaut (commented):
Just a small comment, also i will add a CI check

src/Lightweight/SqlSchema.cpp
christianparpart: true, but i want the constructed column struct to always be valid. This here would now throw in case of failure. I think it should not throw the cryptic but rather an actual "sql column type is unsupported".

hmm... okay, for `ddl2cpp` that case is an unrecoverable error (hence, exception is okay). Do we have any usecase where we want/need to see a database schema where it is okay to have incomplete data, e.g. due to unsupported column types? 🤔

christianparpart: Okay, we're raising an exception now.

PR #187, opened 2025-01-25
Add support for aggregate function calls in SELECT query builder

### LINKED ISSUE #184: Add support for aggregate fields in SELECT query builder
We could get inspired by sqlalchemy's aggregate API.

Functions to be supported:

- `min(fieldName)`
- `max(fieldName)`
- `sum(fieldName)`
- `avg(fieldName)`
- `count(fieldName)`

where `fieldName` is a field to select, such as a table's column name, or a fully qualified name to a specific table (e.g. due to joining multiple tables together, we may need to be explicit there).

The field name could be provided as string literal, or via `&ModelName::FieldName`, and the API implementation itself infers the column name based on the model's name (like we already do in the DataMapper API)

christianparpart:
Closes #184

### DIFF
--- a/src/Lightweight/SqlQuery/Select.cpp
+++ b/src/Lightweight/SqlQuery/Select.cpp
@@ -16,6 +16,7 @@ SqlSelectQueryBuilder& SqlSelectQueryBuilder::Field(std::string_view const& fiel
     m_query.fields += '"';
     m_query.fields += fieldName;
     m_query.fields += '"';
+    m_aliasAllowed = true;
 
     return *this;
 }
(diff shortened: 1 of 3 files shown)


Yaraslaut (commented):
I am not sure about the api for aliases, at the moment we have a function `FieldAs` but here we will have `Field(max("foo").as('f"))` which is significantly different, I think we can merge it as an addition to the library and then update logic for aliasing

Yaraslaut (approved):

christianparpart:
> I am not sure about the api for aliases, at the moment we have a function `FieldAs` but here we will have `Field(max("foo").as('f"))` which is significantly different, I think we can merge it as an addition to the library and then update logic for aliasing

I generally agree here on that point. I was assuming that (for example) `Aggregate::Sum("field").As("alias")` is probably more intuitive to write. Adding `SumAs` to it would multiply the number of function signatures. I wonder if we instead could work on FieldAs() to be something like Field("foo").As("bar").

Yaraslaut:
I think that `Field("foo").As("bar")` is quite convenient and separates field name from the temporary name

christianparpart:
@Yaraslaut FYI: I'm going to implement the suggestion right away in this PR.

Yaraslaut (approved):
LGTM, thanks

PR #190, opened 2025-01-29
Allow generation headers with aliases in dll2cpp

Yaraslaut:


src/tools/ddl2cpp.cpp
```diff
@@ -90,7 +90,17 @@ class CxxModelPrinter
     mutable std::vector<std::string> m_forwwardDeclarations;
     std::stringstream m_definitions;
 
+    struct Config
+    {
+        bool useAliases = false;
+    } m_config;
```
christianparpart, line 96: actually, I was planning to converge to our Contour-style in Lightweight, uing `_config` rather than `m_`-prefixes for private fields :)

src/tools/tests/aliases.custom
```diff
@@ -0,0 +1,17 @@
+CREATE TABLE IF NOT EXISTS User (
```
christianparpart, line 1: can we leave SQL files with their designated file extension? This also helps yntax highlighting. What about `customAliases.sql` or so? - I'm having a `ColumnTypes.sql` locally, that I wanted to add also, but then realized, that not all types work nicely yet (will share with you asap)

src/tools/tests/aliases.custom.expected.cpp
```diff
@@ -0,0 +1,35 @@
+// SPDX-License-Identifier: Apache-2.0
+#pragma once
+#include <Lightweight/DataMapper/DataMapper.hpp>
+#include <Lightweight/SqlConnection.hpp>
+#include <Lightweight/SqlDataBinder.hpp>
+#include <Lightweight/SqlQuery.hpp>
+#include <Lightweight/SqlQueryFormatter.hpp>
+#include <Lightweight/SqlScopedTraceLogger.hpp>
+#include <Lightweight/SqlStatement.hpp>
+#include <Lightweight/SqlTransaction.hpp>
```
christianparpart, line 10: do we wanna clean-up the headers here, while touching it?

src/tools/tests/aliases.custom
Yaraslaut: in test.sh script we check for all `*.sql` files, but in this case we need to add additional parameter to the `ddl2cpp`

src/tools/ddl2cpp.cpp
Yaraslaut: Updated

src/tools/tests/aliases.custom.expected.cpp
Yaraslaut: updated

src/tools/tests/aliases.custom
christianparpart: in the `*.sql` files we could actually greb for `-- processing: custom`, and if such line exists, the custom processing is used rather than the standard processing.

Yaraslaut: hm, so something similar to llvm-lit ?

Yaraslaut: Will update, but tomorrow morning

christianparpart: > hm, so something similar to llvm-lit ?

precisely :D

src/tools/ddl2cpp.cpp
```diff
@@ -87,33 +87,37 @@ std::string MakeVariableName(SqlSchema::FullyQualifiedTableName const& table)
...
     std::string str(std::string_view modelNamespace) const
     {
-        std::ranges::sort(m_forwwardDeclarations);
+        std::ranges::sort(_forwardDeclarations);
 
         std::stringstream output;
         output << "// SPDX-License-Identifier: Apache-2.0\n";
         output << "#pragma once\n";
         output << "#include <Lightweight/DataMapper/DataMapper.hpp>\n";
         output << "#include <Lightweight/SqlConnection.hpp>\n";
```
christianparpart, line 112: do we actually need SqlConnection here?

src/tools/ddl2cpp.cpp
```diff
@@ -133,26 +137,36 @@ class CxxModelPrinter
             cxxPrimaryKeys += '"' + key + '"';
         }
 
-        m_definitions << std::format("struct {} final\n", table.name);
-        m_definitions << std::format("{{\n");
+        _definitions << std::format("struct {} final\n", table.name);
+        _definitions << std::format("{{\n");
```
christianparpart, line 141: We should also add the feature of optionally fill the `TableName` alias iff `_config.useAliases` is set.

src/tools/ddl2cpp.cpp
```diff
@@ -305,6 +320,10 @@ std::variant<Configuration, int> ParseArguments(int argc, char const* argv[])
                 return { EXIT_FAILURE };
             config.outputFileName = argv[i];
         }
+        else if (argv[i] == "--use-aliases"sv)
```
christianparpart, line 323: isn't it more like `--make-aliases` (or create...) rather than `--use-aliases`? :thinking:

src/tools/ddl2cpp.cpp
Yaraslaut: removed

src/tools/ddl2cpp.cpp
Yaraslaut: Updated

src/tools/ddl2cpp.cpp
Yaraslaut: changed

src/tools/ddl2cpp.cpp
```diff
@@ -133,29 +135,61 @@ class CxxModelPrinter
...
         }
 
         for (auto const& foreignKey: table.foreignKeys)
         {
-            m_definitions << std::format(
+            _definitions << std::format(
                 "    BelongsTo<&{}> {};\n",
                 [&]() {
-                    return std::format("{}::{}", foreignKey.primaryKey.table, "id"); // TODO
+                    return std::format("{}::{}", foreignKey.primaryKey.table, foreignKey.primaryKey.columns[0]); // TODO
```
christianparpart, line 192: FYI: `BelongsTo<>` also supports providing an alias name, which we hould (if requested). :)

src/tools/tests/aliases.custom
Yaraslaut: Updated

christianparpart:
@Yaraslaut my brain is about to enter GC-run. Let's discuss the progress here tomorrow morning.

christianparpart (approved):
I f'pushed some little changes into it, but overall LGTM. Thank you. :)

PR #192, opened 2025-01-30
Ensure DataMapper works with models within a namespace

christianparpart:
might be, that we need/want to have models within their own namespace. Hence, we should support it.

@Yaraslaut / @FelixTheC  we should decide on the name inference that we'd like to expect here. Currently the tests are breaking on the struct's name. I think it could be safe to cut off the namespace part and only use the struct's name alone. We might want to extend reflection-cpp for easy access of that information. :thinking: 

EDIT: ok i fixed the test.


Yaraslaut (approved):
LGTM, thanks

PR #194, opened 2025-01-30
Add ability to generically access a record's primary key without knowing its field name

christianparpart:
I'm hijacking this PR as I think it's not agood idea to implement dm-caching right now.

However, it certainly makes sense to keep the helper API that I created, in order to generically find and reference the primary key's field of a record, without knowing its name.

p.s.: see test file for example uses :)


Yaraslaut (approved):
great, thanks

PR #197, opened 2025-02-02
Overall cleanup as suggested by clang-tidy :)

christianparpart:
btw, we still fail, because `std::expected<>` is not available on Github actions's clang-tidy (why?)

### DIFF
--- a/src/Lightweight/SqlError.cpp
+++ b/src/Lightweight/SqlError.cpp
@@ -7,7 +7,7 @@ SqlException::SqlException(SqlErrorInfo info, std::source_location sourceLocatio
     std::runtime_error(std::format("{}", info)),
     _info { std::move(info) }
 {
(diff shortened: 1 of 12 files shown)


src/Lightweight/SqlError.cpp
```diff
@@ -7,7 +7,7 @@ SqlException::SqlException(SqlErrorInfo info, std::source_location sourceLocatio
     std::runtime_error(std::format("{}", info)),
     _info { std::move(info) }
 {
-    SqlLogger::GetLogger().OnError(info, sourceLocation);
+    SqlLogger::GetLogger().OnError(_info, sourceLocation);
```
christianparpart, line 10: this was actually a bug here

Yaraslaut:
Lets update compiler as well, now we are using clang-18

Yaraslaut (approved):
lgtm, thanks :)

Issue #198, opened 2025-02-03
Partially fill entity with DataMapper


FelixTheC:
It would be good to be able to fill an entity via DataMapper only partialy.

```c++
    struct LabList final {
     static constexpr std::string_view TableName = "TABLE"sv;
     Field < int32_t, PrimaryKey::AutoAssign, SqlRealName{ "A" } > a;
     Field < int32_t, PrimaryKey::AutoAssign, SqlRealName{ "B" } > b;
     Field < int32_t, PrimaryKey::AutoAssign, SqlRealName{ "C" } > c; 
     Field < int32_t, PrimaryKey::AutoAssign, SqlRealName{ "D" } > d; 
      Field < int32_t, PrimaryKey::AutoAssign, SqlRealName{ "E" } > e;
    };

  auto const selectQuery = stmt.value().Query("TABLE")
          .Select()
          .Fields({"A"sv,
                   "B"sv,
                   "C"sv
                  }, "TABLE")
          .OrderBy("NR"sv)
          .All();

DataMapper dm{};
std::vector<Entity> results = dm.Query<Entity>(selectQuery )
```

PR #200, opened 2025-02-03
Implement partial read from the database into structure

Yaraslaut:
Closes https://github.com/LASTRADA-Software/Lightweight/issues/198

### DIFF
--- a/src/Lightweight/DataMapper/DataMapper.hpp
+++ b/src/Lightweight/DataMapper/DataMapper.hpp
@@ -18,6 +18,7 @@
 #include <cassert>
 #include <concepts>
 #include <type_traits>
+#include <utility>
 
 /// @defgroup DataMapper Data Mapper
 ///
@@ -45,6 +46,10 @@ constexpr size_t RecordStorageFieldCount =
 
 template <typename Record>
 concept RecordWithStorageFields = (RecordStorageFieldCount<Record> > 0);
+
+/// @brief Represents a sequence of indexes that can be used alongside Query() to retrieve only part of the record.
+template <size_t... Ints>
+using SqlElements = std::integer_sequence<size_t, Ints...>;
 
 namespace detail
 {
@@ -179,6 +184,34 @@ class DataMapper
     template <typename Record, typename... InputParameters>
     std::vector<Record> Query(std::string_view sqlQueryString, InputParameters&&... inputParameters);
 
+    /// Queries records from the database, based on the given query and can be used to retrieve only part of the record
+    /// by specifying the ElementMask.
+    ///
+    /// example:
+    /// @code
+    ///
+    /// struct Person
+    /// {
+    ///    int id;
+    ///    std::string name;
+    ///    std::string email;
+    ///    std::string phone;
+    ///    std::string address;
+    ///    std::string city;
+    ///    std::string country;
+    /// };
+    ///
+    /// auto infos = dm.Query<SqlElements<1,5>(RecordTableName<Person>.Fields({"name"sv, "city"sv}));
+    ///
+    /// for(auto const& info : infos)
+    /// {
+    ///    // only info.name and info.city are loaded
+    /// }
+    /// @endcode
+    template <typename ElementMask, typename Record, typename... InputParameters>
+    std::vector<Record> Query(SqlSelectQueryBuilder::ComposedQuery const& selectQuery,
+                              InputParameters&&... inputParameters);
+
     /// Checks if the record has any modified fields.
     template <typename Record>
     bool IsModified(Record const& record) const noexcept;
@@ -239,6 +272,12 @@ class DataMapper
     void BindOutputColumns(Record& record);
 
     template <typename Record>
+    void BindOutputColumns(Record& record, SqlStatement* stmt);
+
+    template <typename ElementMask, typename Record>
+    void BindOutputColumns(Record& record);
+
+    template <typename ElementMask, typename Record>
     void BindOutputColumns(Record& record, SqlStatement* stmt);
 
     template <size_t FieldIndex, auto ReferencedRecordField, typename Record>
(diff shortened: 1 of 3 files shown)


src/tests/DataMapperTests.cpp
```diff
@@ -669,6 +669,30 @@ TEST_CASE_METHOD(SqlTestFixture, "Query: SELECT into simple struct", "[DataMappe
...
+        person.name = SqlAnsiString<25>(std::format("John-{}", i));
+        person.age = i;
+        dm.Create(person);
+    }
+
+    auto result = dm.Query<SqlElements<1, 3>, Person >(
+        dm.FromTable(RecordTableName<Person>).Select().Fields({ "name"sv, "age"sv }).All());
+
+    for(int i = 20; i<=50; ++i)
+    {
```
christianparpart, line 690:
```suggestion
    {
        CAPTURE(i);
```

src/Lightweight/DataMapper/DataMapper.hpp
```diff
@@ -46,6 +47,9 @@ constexpr size_t RecordStorageFieldCount =
 template <typename Record>
 concept RecordWithStorageFields = (RecordStorageFieldCount<Record> > 0);
 
+template <size_t... Ints>
+using SqlElements = std::integer_sequence<size_t, Ints...>;
```
christianparpart, line 51: Let's give this a `///` documentation as this is meant to be public API :)

src/Lightweight/DataMapper/DataMapper.hpp
```diff
@@ -179,6 +183,12 @@ class DataMapper
     template <typename Record, typename... InputParameters>
     std::vector<Record> Query(std::string_view sqlQueryString, InputParameters&&... inputParameters);
 
+    /// Queries records from the database, based on the given query and can be used to retrieve only part of the record
+    /// by specifying the ElementMask.
```
christianparpart, line 187: I think this definitely deserves a small code example in the comment. What do you think? Can this be trivially shown in a few lines only?

src/Lightweight/DataMapper/DataMapper.hpp
```diff
@@ -670,6 +686,32 @@ std::vector<Record> DataMapper::Query(std::string_view sqlQueryString, InputPara
...
+    auto result = std::vector<Record> {};
+
+    auto record = Record {};
+    BindOutputColumns<ElementMask>(record);
+    ConfigureRelationAutoLoading(record);
+
+    while (_stmt.FetchRow())
+    {
+        result.emplace_back(std::move(record));
+        record = Record {};
```
christianparpart, line 707: After `std::move()`, the moved  value is in a "valid but unspecified state".
Knowing what we're doing with `BindOutputColumns()` and `FetchRow()`, i wonder if we really need to post-re-initialize here again, or if we could actually skip that here.

That'll not add a huge gain to the table, but data is also not appended but actually "assigned" instead. All we care about is that the record is valid, regardless of its element states.

src/tests/DataMapperTests.cpp
```diff
@@ -182,7 +182,7 @@ TEST_CASE_METHOD(SqlTestFixture, "iterate over database", "[SqlRowIterator]")
     for (int i = 40; i <= 50; ++i)
     {
         auto person = Person {};
-        person.name = "John";
+        person.name = SqlAnsiString<25>(std::format("John-{}", i));
```
christianparpart, line 185:
```suggestion
        person.name = std::format("John-{}", i);
```
Looks like we rather want to add a convenience assignment operator to allow assigning an `std::basic_string<T>` to an `SqlFixedString<N, T>`, leaving that line change as intuitive as suggested above. :)

src/tests/DataMapperTests.cpp
```diff
@@ -191,7 +191,7 @@ TEST_CASE_METHOD(SqlTestFixture, "iterate over database", "[SqlRowIterator]")
     int age = 40;
     for (auto&& person: SqlRowIterator<Person>(stmt))
     {
-        CHECK(person.name.Value() == "John");
+        CHECK(person.name.Value() == SqlAnsiString<25>(std::format("John-{}", age)));
```
christianparpart, line 194: same here, for `operator==(std::basic_string<T> const&)`

src/tests/DataMapperTests.cpp
```diff
@@ -669,6 +669,31 @@ TEST_CASE_METHOD(SqlTestFixture, "Query: SELECT into simple struct", "[DataMappe
...
+TEST_CASE_METHOD(SqlTestFixture, "Query: Partial retriaval of the data", "[DataMapper]")
+{
+    auto dm = DataMapper {};
+
+    dm.CreateTable<Person>();
+
+    for (int i = 20; i <= 50; ++i)
+    {
+        auto person = Person {};
+        person.name = SqlAnsiString<25>(std::format("John-{}", i));
```
christianparpart, line 681: ditto (ugh, sorry, I hope I'm not repeating myself lol ) :D

src/tests/DataMapperTests.cpp
```diff
@@ -669,6 +669,31 @@ TEST_CASE_METHOD(SqlTestFixture, "Query: SELECT into simple struct", "[DataMappe
...
+        dm.Create(person);
+    }
+
+    auto result = dm.Query<SqlElements<1, 3>, Person>(
+        dm.FromTable(RecordTableName<Person>).Select().Fields({ "name"sv, "age"sv }).All());
+
+    for (int i = 20; i <= 50; ++i)
+    {
+        CAPTURE(i);
+        CHECK(result[i - 20].name.Value() == SqlAnsiString<25>(std::format("John-{}", i)));
```
christianparpart, line 692: last. WORD.

christianparpart (changes requested):
small ones, but I think we could actually do extend SqlFixedString<> to make things easier to write again in here :)

src/Lightweight/DataMapper/DataMapper.hpp
Yaraslaut: replaced it with `emplace_back()` calls

src/tests/DataMapperTests.cpp
Yaraslaut: this looks like not a very trivial change since person.name here is a field and we need to properly propagate constructors, and test "SqlFixedString: assign" already creates some API that we need to preserve

christianparpart: Rebase on top of hopefully soon merged #202, and then you should be good to go?

christianparpart: this is the only thing missing now, right?

Yaraslaut: Yes

christianparpart (approved):
thank's a lot!

PR #203, opened 2025-02-04
Avoid reusing the same variable name again and allow returning the input Record& ref in BindOutputColumns()

christianparpart:
needed for enable better chaining in
```cpp
std::vector<Record> DataMapper::Query(SqlSelectQueryBuilder::ComposedQuery const // ...
```
for a more reasonable implementation


Yaraslaut (approved):
thanks :)

PR #210, opened 2025-02-06
Fix FieldNameOf for BelongsTo entries with SqlRealName

Yaraslaut:


src/tests/QueryBuilderTests.cpp
```diff
@@ -278,6 +279,28 @@ TEST_CASE_METHOD(SqlTestFixture, "SqlQueryBuilder.FieldsWithBelongsTo", "[SqlQue
 
 }
 
+
+struct QueryBuilderTestEmailWithAliases
+{
+    Field<std::string, SqlRealName { "FAX_ADRESS" }  > email;
+    BelongsTo<&UsersFields::name, SqlRealName{ "UUID"}> user;
```
christianparpart, line 286:
```suggestion
    Field<std::string, SqlRealName { "FAX_ADRESS" }> email;
    BelongsTo<&UsersFields::name, SqlRealName{ "USER_REC" }> user;
```

"UUID" is not the best column name in an SQL table, as some might have it as a keyword. 
But actually, I wanted to suggest proper formatting

christianparpart (changes requested):
formatting should be complained about, I wonder. :thinking:

src/tests/QueryBuilderTests.cpp
Yaraslaut: Updated it, and also clang-format was ok with this formatting xD

christianparpart:
oops. should I have mentioned, that the tests needs adjusting accordingly, as well? :innocent:

christianparpart (approved):
yay. thanks for fixing it.

src/tests/QueryBuilderTests.cpp
christianparpart: This IS weird. Interesting!

PR #212, opened 2025-02-08
Add QuerySparse<P...>()

christianparpart:
WIP


.github/workflows/build.yml
```diff
@@ -263,8 +259,10 @@ jobs:
       - name: "Extract package"
         run: |
           sudo tar -xvf Lightweight-Linux-CI.tar.gz --strip-components=1 -C /usr/local
+      - name: "Update APT database"
+        run: sudo apt -q update
```
christianparpart, line 263: the only way to get Ubuntu images green was with this damn line. Too bad. ;(

src/Lightweight/DataMapper/DataMapper.hpp
```diff
@@ -68,6 +68,239 @@ auto ToSharedPtrList(Container<Object, Allocator<Object>> container)
...
+                                                   _searchCondition.tableAlias,
+                                                   _searchCondition.tableJoins,
+                                                   _searchCondition.condition,
+                                                   orderBy,
+                                                   n));
+        Derived::ReadResults(_stmt.GetResultCursor(), &records);
+        return records;
+    }
+
+    [[nodiscard]] std::vector<Record> Range(size_t offset, size_t limit)
```
Yaraslaut, line 139: I think we need to rename this function, this is confusing, Range should accept start and end, not start and number of elements

Yaraslaut (approved):

PR #217, opened 2025-02-11
Fix SqlRowIterator and also make it more efficient

christianparpart:
@Yaraslaut it wasn't actually working as expected. It always only fetched once, and that wasn't working neither :)
The test case didn't catch it, because it was not testing how many rows it has fetched (or not).

This should be fixing it, and make it also slightly more efficient.


Yaraslaut (approved):
Lgtm, many thanks

christianparpart:
Thank you very much, for finishing it, @Yaraslaut :)

PR #221, opened 2025-02-14
[logging] Make sure logging is thread safe

### LINKED ISSUE #220: make read/write access to `SqlStandardLogger::currentTimeStr` thread-safe
we must expect to log SQL from multiple threads concurrently. Therefore it makes sense to make the logger thread safe, specifically the access to the string that holds the current time (read and write access).

christianparpart:
closes #220.


Yaraslaut (approved):
thanks :)

PR #228, opened 2025-03-07
aim to help easing up bridging between modern SQL code and legacy variant-list based model architectures

christianparpart:
this aims to add some convenience methods to ease (and demo) the use of trivially transforming a compile time reflected data model from/to a dynamically grown and of type unknown column list (e.g. `SqlVariantRow`).

with this, I realized that `SqlVariant` did not support `SqlGuid` just yet.


src/Lightweight/DataBinder/SqlVariant.hpp
```diff
@@ -297,13 +332,52 @@ struct SqlVariant
...
+    Reflection::EnumerateMembers(record, [&]<size_t I>(auto& field) mutable {
+        using ValueType = typename Reflection::MemberTypeOf<I, Record>::ValueType;
+        // TODO: care about NULL columns
+        // TODO: care about type conversions (SqlFixedString<> etc)
+        field = row[I].Get<ValueType>();
+    });
+
+    return record;
+}
+
```
Yaraslaut, line 380: I do not see where we use it, also no tests for this one

src/Lightweight/DataBinder/SqlVariant.hpp
```diff
@@ -297,13 +332,52 @@ struct SqlVariant
...
+    {
+        if (IsNull())
+            return std::nullopt;
+
+        if (auto const* guid = std::get_if<SqlGuid>(&value))
+            return *guid;
+
+        throw std::bad_variant_access();
+    }
+
```
Yaraslaut, line 346: please add test for this one, including exception throw

src/Lightweight/DataBinder/SqlVariant.hpp
```diff
@@ -265,6 +275,31 @@ struct SqlVariant
...
+        return ToString() <=> other.ToString();
+    }
+
+    template <typename T>
+    std::weak_ordering operator<=>(std::optional<T> const& other) const noexcept
+    {
+        if (IsNull())
+            return other.has_value() ? std::weak_ordering::less : std::weak_ordering::equivalent;
+        return ToString() <=> SqlVariant { other }.ToString();
+    }
```
Yaraslaut, line 289: why do we need it?

Yaraslaut (changes requested):
We need to add some tests and i also have some small comments

src/Lightweight/DataBinder/SqlVariant.hpp
christianparpart: yeah there was a test until I realized that this is not memory-efficient in our use-case internally. We could drop these again. The other one in Record.hpp is more important & efficient.

src/Lightweight/DataBinder/SqlVariant.hpp
christianparpart: Thanks, it actually caught a missing implementation. lol.

src/Lightweight/DataBinder/SqlVariant.hpp
christianparpart: not anymore. I've dropped it. I've adapted the test cases to complicitly compare SqlVariant with SqlVariant now. Better be explicit ;)

Yaraslaut (approved):
lgtm, thanks :)

Issue #231, opened 2025-03-31
Use ISO8601 for the sql date time formatting


Yaraslaut:
Use the following formatting : https://www.iso.org/iso-8601-date-and-time-format.html

christianparpart:
@Yaraslaut: you mean for `std::formatter<SqlDateTime>`?

PR #232, opened 2025-04-04
Ensure that std::formatter for sqldatetime is in iso standard

Yaraslaut:
Closes https://github.com/LASTRADA-Software/Lightweight/issues/231

### DIFF
--- a/src/Lightweight/DataBinder/SqlDateTime.hpp
+++ b/src/Lightweight/DataBinder/SqlDateTime.hpp
@@ -172,14 +172,19 @@ struct std::formatter<SqlDateTime>: std::formatter<std::string>
     LIGHTWEIGHT_FORCE_INLINE auto format(SqlDateTime const& value, std::format_context& ctx) const
         -> std::format_context::iterator
     {
-        return std::formatter<std::string>::format(std::format("{:04}-{:02}-{:02} {:02}:{:02}:{:02}.{:09}",
+        // This can be used manually to format the date and time inside sql query builder,
+        // that is why we need to use iso standard 8601, also millisecond precision is used for the formatting
+        //
+        // internally fraction is stored in nanoseconds, here we need to get it in milliseconds
+        const auto milliseconds = value.sqlValue.fraction / 1'000'000;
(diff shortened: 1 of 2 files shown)


src/Lightweight/DataBinder/SqlDateTime.hpp
```diff
@@ -172,14 +172,19 @@ struct std::formatter<SqlDateTime>: std::formatter<std::string>
     LIGHTWEIGHT_FORCE_INLINE auto format(SqlDateTime const& value, std::format_context& ctx) const
         -> std::format_context::iterator
     {
-        return std::formatter<std::string>::format(std::format("{:04}-{:02}-{:02} {:02}:{:02}:{:02}.{:09}",
+        // This can be used manually to format the date and time inside sql query builder,
+        // that is why we need to use iso standard 8601, also millisecond precision is used for the formatting
+        //
+        // internally fraction is stored in nanoseconds, here we need to get it in milliseconds
+        const auto milliseconds = value.sqlValue.fraction / 1'000'000;
```
christianparpart, line 179:
```suggestion
        using namespace std::chrono;
        const auto milliseconds = duration_cast<milliseconds>(nanoseconds(value.sqlValue.fraction));
```

would that make it more readable?

src/tests/QueryBuilderTests.cpp
```diff
@@ -1178,3 +1179,14 @@ TEST_CASE_METHOD(SqlTestFixture, "AlterTable AddForeignKeyColumn", "[SqlQueryBui
                     )sql",
         });
 }
+
+TEST_CASE_METHOD(SqlTestFixture, "SqlQueryBuilder: SqlDateTime formatting", "[SqlQueryBuilder]")
+{
+
+    const auto tp_micros = time_point_cast<std::chrono::milliseconds>(std::chrono::system_clock::now());
```
christianparpart, line 1186: don't you think this is a little non-deterministic? If we'd use a fixed timestamp here, we could check aganist a hard string literal below and not depnd on std::format. :thinking:

christianparpart (approved):

PR #233, opened 2025-04-07
[dd2lcpp] Allow configuration of member name formatting through cli

Yaraslaut:
Added arguments to the dd2cpp to format members according to some formatting, implemented:
 * `camelCase` 
 *  `snake_case`

### DIFF
--- a/src/tools/ddl2cpp.cpp
+++ b/src/tools/ddl2cpp.cpp
@@ -2,6 +2,7 @@
 #include <Lightweight/SqlConnection.hpp>
 #include <Lightweight/SqlSchema.hpp>
 #include <Lightweight/SqlStatement.hpp>
+#include <Lightweight/Utils.hpp>
 
 #include <algorithm>
 #include <cassert>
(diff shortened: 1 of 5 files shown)


src/tools/ddl2cpp.cpp
```diff
@@ -83,15 +83,96 @@ std::string MakeVariableName(SqlSchema::FullyQualifiedTableName const& table)
     return name;
 }
 
+enum class FormatType : uint8_t
+{
+    snake_case,
+    camelCase,
+    doNotChange,
```
christianparpart, line 90: I think we should not encode the casing into the name itself but rather have it as a comment above / next to it.

Also, doNotChange is the only one that contains the action "do" in it, which it shouldn't, IMHO.

src/tools/ddl2cpp.cpp
```diff
@@ -83,15 +83,96 @@ std::string MakeVariableName(SqlSchema::FullyQualifiedTableName const& table)
...
+};
+
+std::string FormatName(std::string_view name, FormatType formatType)
```
christianparpart, line 93: this sounds like we definitely want test cases for it. we could move this function out of ddl2cpp.cpp (e.g. into Utils), to have it testable.

christianparpart: NB: function names are camelCase, whereas type names are PascalCase. :)

Yaraslaut: as far as i see, function names are PascalCase, we should add this into clang-tidy check

christianparpart:
you ve got me. ;)

christianparpart (approved):
Thank you, @Yaraslaut

PR #234, opened 2025-04-08
Generate example from dd2cpp

Yaraslaut:
Generate example from ddl2cpp and create cmake target to generate executable 
Also for long var char with size 0 use `SqlDynamicAnsiString`


src/tools/ddl2cpp.cpp
```diff
@@ -72,7 +72,11 @@ std::string MakeType(SqlSchema::Column const& column)
                        // TODO distinguish between BINARY and VARBINARY
                        // (https://github.com/LASTRADA-Software/Lightweight/issues/182)
                        [](VarBinary const& type) -> std::string { return std::format("SqlBinary<{}>", type.size); },
-                       [](Varchar const& type) -> std::string { return std::format("SqlAnsiString<{}>", type.size); },
+                       [](Varchar const& type) -> std::string {
+                           if (type.size > 0)
```
christianparpart, line 76: (x > 0 && x < 4000)

I think we have a symbolic constant to represent that literal 4000, too.

src/tools/ddl2cpp.cpp
```diff
@@ -72,7 +72,11 @@ std::string MakeType(SqlSchema::Column const& column)
...
+                               return std::format("SqlAnsiString<{}>", type.size);
+                           return std::format("SqlDynamicAnsiString<{}>", 255); // put something else ?
```
christianparpart, line 78: yeah, the 255 should be the actual value, if non-zero :)

src/tools/ddl2cpp.cpp
```diff
@@ -141,6 +146,7 @@ class CxxModelPrinter
         output << "// SPDX-License-Identifier: Apache-2.0\n";
         output << "#pragma once\n";
         output << "#include <Lightweight/DataMapper/DataMapper.hpp>\n";
+        output << "using namespace std::string_view_literals;\n";
```
christianparpart, line 149:
```suggestion
        output << '\n';
        output << "using namespace std::string_view_literals;\n";
```
We should also have a separation line between pragma once and the include section, btw

src/tools/ddl2cpp.cpp
```diff
@@ -155,6 +161,44 @@ class CxxModelPrinter
...
+        };
+
+        const auto tableName = aliasTableName(table.name);
+        exampleEntries << std::format("auto entries{} = dm.Query<{}>().All();\n", tableName, tableName);
+
+        exampleEntries << std::format("for (auto const& entry: entries{})\n", tableName);
+        exampleEntries << "{\n";
+
+        const auto column = table.columns[0];
+        const auto memberName = formatName(column.name, _config.formatType);
```
christianparpart, line 182: isn't that FormatName now? 🤔

src/tools/ddl2cpp.cpp
```diff
@@ -373,6 +418,10 @@ std::variant<Configuration, int> ParseArguments(int argc, char const* argv[])
             config.outputFileName = argv[i];
             std::println("Output file name: {}", config.outputFileName);
         }
+        else if (argv[i] == "--generate-example"sv)
```
christianparpart, line 421: Not in the README.md

PR #237, opened 2025-04-12
breakout PR from dbtool WIP PR

christianparpart:
some things we can merge early, such that we won't accidentally do the same or create merge conflicts along the lines

### DIFF
--- a/src/Lightweight/SqlColumnTypeDefinitions.hpp
+++ b/src/Lightweight/SqlColumnTypeDefinitions.hpp
@@ -2,7 +2,11 @@
 
 #pragma once
 
+#include <optional>
 #include <variant>
+
+#include <sql.h>
+#include <sqlext.h>
 
 namespace SqlColumnTypeDefinitions
 {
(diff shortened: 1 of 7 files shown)


src/Lightweight/SqlColumnTypeDefinitions.hpp
```diff
@@ -50,3 +58,48 @@ using SqlColumnTypeDefinition = std::variant<SqlColumnTypeDefinitions::Bigint,
...
+    using namespace SqlColumnTypeDefinitions;
+    // clang-format off
+    switch (value)
+    {
+        case SQL_BIGINT: return Bigint {};
+        case SQL_BINARY: return Binary { size };
+        case SQL_BIT: return Bool {};
+        case SQL_CHAR: return Char { size };
+        case SQL_DATE: return Date {};
+        case SQL_DECIMAL: return Decimal { .precision = precision, .scale = size };
```
Yaraslaut, line 80: assert is missing here now

src/Lightweight/SqlColumnTypeDefinitions.hpp
```diff
@@ -50,3 +58,48 @@ using SqlColumnTypeDefinition = std::variant<SqlColumnTypeDefinitions::Bigint,
...
+        case SQL_DOUBLE: return Real {};
+        case SQL_FLOAT: return Real {};
+        case SQL_GUID: return Guid {};
+        case SQL_INTEGER: return Integer {};
+        case SQL_LONGVARBINARY: return VarBinary { size };
+        case SQL_LONGVARCHAR: return Varchar { size };
+        case SQL_NUMERIC: return Decimal { .precision = precision, .scale = size };
```
Yaraslaut, line 87: assert here as well

Yaraslaut (approved):
LGTM, thanks just a small thing to consider is to return assert in the future

src/Lightweight/SqlColumnTypeDefinitions.hpp
christianparpart: in both cases, assert won't work anymore, because it's `constexpr` now. Also, the assert was hitting in our code (you remember?)... We may need to re-read the docs again first (we can chat about it tomorrow) :)

PR #238, opened 2025-04-13
[ddl2cpp] handle BelongsTo generation better

Yaraslaut:


src/tools/ddl2cpp.cpp
```diff
@@ -69,9 +70,11 @@ std::string MakeType(SqlSchema::Column const& column)
                        [](Time const&) -> std::string { return "SqlTime"; },
                        [](Timestamp const&) -> std::string { return "SqlDateTime"; },
                        [](Tinyint const&) -> std::string { return "uint8_t"; },
-                       // TODO distinguish between BINARY and VARBINARY
+                       // TODO distinguish between BINARY and VARBINARY and VARBINARY(MAX) which is for an IMAGE
```
christianparpart, line 73: we can already provide a `Binary` next to `VarBinary` to properly stub this out here

src/tools/ddl2cpp.cpp
```diff
@@ -478,6 +500,50 @@ std::variant<Configuration, int> ParseArguments(int argc, char const* argv[])
     return { config };
 }
 
+void resolveOrderAndPrintTable(auto& printer, const auto& tables)
+{
+    std::println("Starting to print tables");
+    std::unordered_map<size_t, int> numberOfForeignKeys;
+    for (auto const& [index, table]: std::views::enumerate(tables))
```
christianparpart, line 507: what was it with `enumerate`? No need to cref, right? just `auto const` :)

christianparpart (approved):
Thank you, @Yaraslaut :)

PR #242, opened 2025-04-23
SqlQuery: Fix escaping on WhereIn() values

### LINKED ISSUE #241: SqlQuery's "WhereIn" for a set of strings is not quoting the string literals
SqlQuery's "WhereIn" for a set of strings is not quoting the string literals

christianparpart:
Fixes #241


Yaraslaut (approved):
many thanks :)

PR #243, opened 2025-04-23
[ddl2cpp] Generate structs into separate files

Yaraslaut:
Create header file for each table

### DIFF
--- a/src/tools/ddl2cpp.cpp
+++ b/src/tools/ddl2cpp.cpp
@@ -7,7 +7,9 @@
 #include <algorithm>
 #include <cassert>
 #include <fstream>
+#include <ostream>
 #include <print>
+#include <sstream>
 #include <unordered_map>
 
 // TODO: have an OdbcConnectionString API to help compose/decompose connection settings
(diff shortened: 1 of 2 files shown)


src/tools/ddl2cpp.cpp
```diff
@@ -134,24 +140,68 @@ class CxxModelPrinter
...
+    void printToFiles(std::string_view modelNamespace, std::string_view outputDirectory)
+    {
+
+        for (auto const& [tableName, definition]: _definitions)
+        {
+            const auto fileName = std::format("{}/{}.hpp", outputDirectory, tableName);
+            auto file = std::ofstream(fileName);
+            if (!file)
+            {
+                std::println("Failed to open file: {}", fileName);
```
christianparpart, line 174: That's more like `("Failed to create file {}. {}", fileName, reason)`

src/tools/ddl2cpp.cpp
```diff
@@ -134,24 +140,68 @@ class CxxModelPrinter
...
+                continue;
+            }
+            file << strTable(modelNamespace, tableName);
+        }
+    }
+
+    std::string strTable(std::string_view modelNamespace,
```
christianparpart, line 181: I'm sorry. I'm having a hard time imagining what this function is doing, when reading this name, `strTable` :)

christianparpart (approved):
drop ` std::cout << std::flush`, then LGTM. many thanks :)

Issue #244, opened 2025-04-24
Add support for querying column data from joined tables


christianparpart:
Given the following basic SQL JOIN syntax:
```sql
SELECT A.*, C.* FROM A
JOIN B ON B.a_id = A.id
JOIN C ON C.id = B.c_id;
```

It would be niceto get that queried via the data mapper API like so:

```cpp
dm.Query<A, C>()
  .InnerJoin<&B::a_id, &A::id>()
  .InnerJoin<&C::b_id, &B::id>()
  .All();
```

Alternatively, slightly more verbose:

```cpp
dm.Query<A, C>()
  .InnerJoin<B, &B::a_id, &A::id>()
  .InnerJoin<C, &C::b_id, &B::id>()
  .All();
```

PR #246, opened 2025-04-24
Add Support for querying column data from joined tables

Yaraslaut:
Refs https://github.com/LASTRADA-Software/Lightweight/issues/244

### DIFF
--- a/src/Lightweight/SqlQuery/Core.hpp
+++ b/src/Lightweight/SqlQuery/Core.hpp
@@ -83,7 +83,7 @@ std::string MakeEscapedSqlString(T const& value)
     std::string escapedValue;
     escapedValue += '\'';
 
-    for (auto const ch : value)
+    for (auto const ch: value)
     {
         // In SQL strings, single quotes are escaped by doubling them.
         if (ch == '\'')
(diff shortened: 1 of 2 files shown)


src/Lightweight/SqlQuery/Core.hpp
```diff
@@ -262,6 +262,14 @@ class [[nodiscard]] SqlWhereClauseBuilder
         requires std::invocable<OnChainCallable, SqlJoinConditionBuilder>
     [[nodiscard]] Derived& InnerJoin(std::string_view joinTable, OnChainCallable const& onClauseBuilder);
 
+    /// Constructs an INNER JOIN clause given two fields from different records
+    /// using the field name as join column.
+    /// Example:
+    /// @code
+    /// InnerJoin<&JoinTestB::a_id, &JoinTestA::id>()
+    template <auto LeftField, auto RightField>
```
christianparpart, line 270:
```suggestion
    /// Constructs an `INNER JOIN` clause given two fields from different records
    /// using the field name as join column.
    /// 
    /// @tparam LeftField  The field name to join on, such as `JoinTestB::a_id`, which will join on table `JoinTestB` with the column `a_id` to be compared against right field's column.
    /// @tparam RightField The other column to compare and join against.
    ///
    /// Example:
    /// @code
    /// InnerJoin<&JoinTestB::a_id, &JoinTestA::id>()
    /// // This will generate a INNER JOIN "JoinTestB" ON "InnerTestB"."a_id" = "JoinTestA"."id"
    template <auto LeftField, auto RightField>
```

christianparpart (approved):

PR #247, opened 2025-04-25
Add main <Lightweight/Lightweight.hpp> header file for public easy use and add missing SPDX header to ddl2cpp.cpp

christianparpart:


Yaraslaut (approved):
Thanks :)

PR #249, opened 2025-04-27
Implement data mapper handling of join statements

Yaraslaut:
Refs https://github.com/LASTRADA-Software/Lightweight/issues/244

This PR finished implementation of data mapper support of join statements

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -97,7 +97,7 @@ jobs:
             "windows-clangcl-debug",
           ]
     name: "Windows-${{ matrix.preset }}"
-    runs-on: windows-latest
+    runs-on: windows-2025
     steps:
       - uses: actions/checkout@v4
       - name: Run sccache-cache
--- a/src/Lightweight/DataMapper/DataMapper.hpp
+++ b/src/Lightweight/DataMapper/DataMapper.hpp
@@ -31,10 +31,12 @@
 /// @ingroup DataMapper
 template <typename T>
 concept FieldWithStorage = requires(T const& field, T& mutableField) {
+    // clang-format off
     { field.Value() } -> std::convertible_to<typename T::ValueType const&>;
     { mutableField.MutableValue() } -> std::convertible_to<typename T::ValueType&>;
     { field.IsModified() } -> std::convertible_to<bool>;
     { mutableField.SetModified(bool {}) } -> std::convertible_to<void>;
+    // clang-format on
 };
 
 /// Represents the number of fields with storage in a record.
@@ -51,12 +53,6 @@ constexpr size_t RecordStorageFieldCount =
 
 template <typename Record>
 concept RecordWithStorageFields = (RecordStorageFieldCount<Record> > 0);
-
-/// @brief Represents a sequence of indexes that can be used alongside Query() to retrieve only part of the record.
-///
-/// @ingroup DataMapper
-template <size_t... Ints>
-using SqlElements = std::integer_sequence<size_t, Ints...>;
 
 namespace detail
 {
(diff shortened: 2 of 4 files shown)


.github/workflows/build.yml
```diff
@@ -97,7 +97,7 @@ jobs:
             "windows-clangcl-debug",
           ]
     name: "Windows-${{ matrix.preset }}"
-    runs-on: windows-latest
+    runs-on: windows-2025
```
christianparpart, line 100: how come? What was the motivation here? (as much as I like windows server 2025 :D)

src/Lightweight/CMakeLists.txt
```diff
@@ -127,6 +127,8 @@ if(MSVC)
     # clang-cl does not support /MP
     if(NOT CLANG_CL)
         target_compile_options(Lightweight PRIVATE /MP)
+    else()
+        target_compile_options(Lightweight PRIVATE /clang:-Wno-unused-lambda-capture)
```
christianparpart, line 131: Why is this a problem? Can't we just get rid of unused lambda captures? I think it's a bad habbit to linger unused captures around. Or what am I missing here?

src/Lightweight/DataMapper/DataMapper.hpp
```diff
@@ -31,11 +31,13 @@
 /// @ingroup DataMapper
 template <typename T>
 concept FieldWithStorage = requires(T const& field, T& mutableField) {
+                               // clang-format off
```
christianparpart, line 34: Weird indentation?

src/Lightweight/DataMapper/DataMapper.hpp
```diff
@@ -599,13 +595,43 @@ class DataMapper
     ///    // only info.name and info.city are loaded
     /// }
     /// @endcode
-    template <typename ElementMask, typename Record, typename... InputParameters>
+    template <typename lementMask, typename Record, typename... InputParameters>
```
christianparpart, line 598: `lementMask`? Please tell me you just accidentally erased that `E` in the beginning. :D

src/Lightweight/DataMapper/DataMapper.hpp
```diff
@@ -599,13 +595,43 @@ class DataMapper
...
+    /// struct JointC{};
+    ///
+    /// // the following query will construct statement to fetch all elements of JointA and JointC types
+    /// auto dm = DataMapper {};
+    /// const auto query = dm.FromTable(RecordTableName<JoinTestA>)
+    ///             .Select()
+    ///             .Fields<JointA, JointC>()
+    ///             .InnerJoin<&JointB::a_id, &JointA::id>()
+    ///             .InnerJoin<&JointC::id, &JointB::c_id>()
+    ///             .All();
```
christianparpart, line 620:
```cpp
auto const result = dm.Query<A, C>()
                      .InnerJoin<&B::a_id, &A::id>()
                      .InnerJoin<&C::b_id, &B::id>()
                      .All();
```
And that's not possible? 🤔 We should see if (while the above should work of course)... if we can get the usage as trivial as possible, for the coder. What do you think?

src/Lightweight/DataMapper/DataMapper.hpp
```diff
@@ -599,13 +595,43 @@ class DataMapper
...
+    /// const auto records= dm.Query<JointA, JointC>(query);
```
christianparpart, line 621:
```suggestion
    /// const auto records = dm.Query<JointA, JointC>(query);
```
Ok, lol. My vision scope today is below zero, lol. But at least let's have proper formatting :D
And also suggest slightly more real-world examples, like:

```suggestion
    /// const auto records = dm.Query<JointA, JointC>().Where....All(); // with correct calls ofc
```

src/Lightweight/DataMapper/DataMapper.hpp
```diff
@@ -1142,6 +1170,40 @@ std::vector<Record> DataMapper::Query(std::string_view sqlQueryString, InputPara
...
+    const auto ConfigureFetchAndBind = [this](auto& record) {
+        auto& [recordFirst, recordSecond] = record;
+        BindOutputColumns<FirstRecord, 1>(recordFirst);
+        BindOutputColumns<SecondRecord, Reflection::CountMembers<FirstRecord> + 1>(recordSecond);
+        ConfigureRelationAutoLoading(recordFirst);
+        ConfigureRelationAutoLoading(recordSecond);
+    };
+
+    auto& firstRecord = result.emplace_back();
+    ConfigureFetchAndBind(firstRecord);
```
christianparpart, line 1192:
```suggestion
    ConfigureFetchAndBind(result.emplace_back());
```

src/Lightweight/DataMapper/DataMapper.hpp
```diff
@@ -1142,6 +1170,40 @@ std::vector<Record> DataMapper::Query(std::string_view sqlQueryString, InputPara
...
+
+    while (_stmt.FetchRow())
+    {
+        auto& record = result.emplace_back();
+        ConfigureFetchAndBind(record);
+    }
```
christianparpart, line 1198:
```suggestion
        ConfigureFetchAndBind(result.emplace_back());
```
I am not sure, but I think there were cases where pre-binding did not work. We may need to take care of that. We do that in other places already, we can talk tomorrow (Monday) in person about it. 🤔

src/Lightweight/DataMapper/Record.hpp
```diff
@@ -10,14 +10,37 @@
...
 
+/// @brief Represents a sequence of indexes that can be used alongside Query() to retrieve only part of the record.
+///
+/// @ingroup DataMapper
+template <size_t... Ints>
+using SqlElements = std::integer_sequence<size_t, Ints...>;
+
+// Helper trait to detect specializations of SqlElements
+template <typename T>
+struct isSqlElements: std::false_type
```
christianparpart, line 21: why camelBack rather than CamelCase? (also below) - if it's meant to be private, let's put it into the detail sub namespace.

.github/workflows/build.yml
Yaraslaut: i wanted to check if false-positive error on windows will disappear with newer version of `clang-cl`

src/Lightweight/CMakeLists.txt
Yaraslaut: removed it

src/Lightweight/DataMapper/DataMapper.hpp
Yaraslaut: fixed formatting, local clang-format was not very good with concepts

src/Lightweight/DataMapper/DataMapper.hpp
Yaraslaut: rvalue -> lvalu& will fail so this is not possible

src/Lightweight/DataMapper/Record.hpp
Yaraslaut: moved into `detail` namespace

PR #254, opened 2025-04-28
ddl2cpp: Implement better primary key assignment declaration

### LINKED ISSUE #252: ddl2cpp: created PKs should not hard-code to "ServerSideAutoIncrement"
First and foremost, some schema's might have fully switched to manual PK value creation, even for integers.

But keep in mind, that there MAY be primary keys that are non-integer, e.g. GUID or VARCHAR. We will need to be able to deal with these as well.

if a server-side setting is not inferred from the Schema, we need it configurable (and at least overridable) from the yaml configuration / CLI flags.

christianparpart:
Closes #252.

Example ddl2cpp.yml now:

```yaml
ConnectionString: 'DSN=YourDSN;UID=YourUser;PWD=YourPassword;TIMEOUT=5'
OutputDirectory: 'src/entities'
MakeAliases: true
NamingConvention: CamelCase
ForceUnicodeTextColumns: true
PrimaryKeyAssignment: ClientSide
```

### DIFF
--- a/src/tools/ddl2cpp.cpp
+++ b/src/tools/ddl2cpp.cpp
@@ -1,5 +1,6 @@
 // SPDX-License-Identifier: Apache-2.0
 
+#include <Lightweight/DataMapper/Field.hpp>
 #include <Lightweight/SqlConnectInfo.hpp>
 #include <Lightweight/SqlConnection.hpp>
 #include <Lightweight/SqlSchema.hpp>
@@ -152,6 +153,7 @@ class CxxModelPrinter
     {
         bool makeAliases = false;
         FormatType formatType = FormatType::camelCase;
+        PrimaryKey primaryKeyAssignment = PrimaryKey::ServerSideAutoIncrement;
         bool forceUnicodeTextColumns = false;
         bool generateExample = false;
     };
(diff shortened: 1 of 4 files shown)


src/tools/ddl2cpp.cpp
```diff
@@ -318,8 +320,13 @@ class CxxModelPrinter
...
             if (column.isPrimaryKey)
             {
+                auto primaryKeyAssignmentText = std::string {};
+                if (_config.primaryKeyAssignment == PrimaryKey::ServerSideAutoIncrement)
+                    primaryKeyAssignmentText = ", PrimaryKey::ServerSideAutoIncrement";
+                else if (_config.primaryKeyAssignment == PrimaryKey::AutoAssign)
+                    primaryKeyAssignmentText = ", PrimaryKey::AutoAssign";
                 definition.text << std::format(
-                    "    Field<{}, PrimaryKey::ServerSideAutoIncrement{}> {};\n", type, aliasName(column.name), memberName);
+                    "    Field<{}{}{}> {};\n", type, primaryKeyAssignmentText, aliasName(column.name), memberName);
```
Yaraslaut, line 329: In the column type we have filed `bool isAutoIncrement = false;` we can use it if no config value is specified, also maybe create lambda to capture the logic of this part of the string, similar to `aliasName` function

PR #255, opened 2025-04-28
Add API to create join query using datamapper

Yaraslaut:
Closes https://github.com/LASTRADA-Software/Lightweight/issues/244

Final part of the handling for join statement

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -22,7 +22,7 @@ jobs:
       - name: update apt
         run: sudo apt -q update
       - name: install dependencies
-        run: sudo apt install -y cmake ninja-build catch2 unixodbc-dev sqlite3 libsqlite3-dev libyaml-cpp-dev libsqliteodbc valgrind uuid-dev g++-14  doxygen
+        run: sudo apt install -y cmake ninja-build catch2 unixodbc-dev sqlite3 libsqlite3-dev libsqliteodbc valgrind uuid-dev g++-14 libyaml-cpp-dev doxygen
       - name: cmake configure
         run: cmake --preset linux-gcc-release -D LIGHTWEIGHT_BUILD_DOCUMENTATION=ON
       - name: build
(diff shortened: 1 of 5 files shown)


.github/workflows/docs.yml
```diff
@@ -22,7 +22,7 @@ jobs:
       - name: update apt
         run: sudo apt -q update
       - name: install dependencies
-        run: sudo apt install -y cmake ninja-build catch2 unixodbc-dev sqlite3 libsqlite3-dev libsqliteodbc valgrind uuid-dev g++-14  doxygen
+        run: sudo apt install -y cmake ninja-build catch2 unixodbc-dev sqlite3 libsqlite3-dev libsqliteodbc valgrind uuid-dev g++-14 libyaml-cpp-dev doxygen
```
christianparpart, line 25: uff, sounds like we both did it :D

src/tests/DataMapperTests.cpp
```diff
@@ -1348,20 +1348,15 @@ TEST_CASE_METHOD(SqlTestFixture, "MapForJointStatement", "[DataMapper]")
...
     }
 
-    auto const records = dm.Query<JoinA, JoinC>(dm.FromTable(RecordTableName<JoinA>)
-                                                    .Select()
-                                                    .Fields<JoinA, JoinC>()
-                                                    .InnerJoin<&JoinB::a_id, &JoinA::id>()
-                                                    .InnerJoin<&JoinC::id, &JoinB::c_id>()
-                                                    .All());
+    auto const records =
+        dm.Query<JoinA, JoinC>().InnerJoin<&JoinB::a_id, &JoinA::id>().InnerJoin<&JoinC::id, &JoinB::c_id>().All();
```
christianparpart, line 1359: :heart: - but do we really wanna adapt this test case? Why not creating a new one. The original one checks query builder, in fact, and the second one operates directly on the DataMapper API.

NB: I think we should have a test for the query-builder-only API in the QueryBuilderTests.cpp instead, :thinking: 

Let's talk to morrow morning in face2face about it.

Yaraslaut: we have separate test for query builder to construct joint

PR #256, opened 2025-04-28
Add handling of constraints that are both primary and foreign keys

Yaraslaut:
Closes https://github.com/LASTRADA-Software/Lightweight/issues/252

### DIFF
--- a/src/tools/ddl2cpp.cpp
+++ b/src/tools/ddl2cpp.cpp
@@ -151,6 +151,7 @@ class CxxModelPrinter
   public:
     struct Config
     {
+        std::string foreignKeyCollisionPrefix;
         bool makeAliases = false;
         FormatType formatType = FormatType::camelCase;
         PrimaryKey primaryKeyAssignment = PrimaryKey::ServerSideAutoIncrement;
@@ -222,7 +223,7 @@ class CxxModelPrinter
         output << "\n";
 
         auto requiredTables = _definitions[tableName].requiredTables;
-        std::sort(requiredTables.begin(), requiredTables.end());
+        std::ranges::sort(requiredTables);
         for (auto const& requiredTable: requiredTables)
             output << std::format("#include \"{}.hpp\"\n", requiredTable);
 
@@ -277,6 +278,8 @@ class CxxModelPrinter
     {
         using namespace std::string_view_literals;
 
+
+        std::vector<std::string> renameForeignKeys;
         auto& definition = _definitions[table.name];
         std::string cxxPrimaryKeys;
         for (auto const& key: table.primaryKeys)
@@ -328,6 +331,12 @@ class CxxModelPrinter
         {
             std::string type = MakeType(column, _config.forceUnicodeTextColumns);
             const auto memberName = FormatName(column.name, _config.formatType);
+
+            if (column.isForeignKey && column.isPrimaryKey)
+            {
+                std::println("Column is both a foreign key and a primary key: {}.{}", table.name, column.name);
+                renameForeignKeys.push_back(column.name);
+            }
             // all foreign keys will be handled in the BelongsTo
             if (column.isPrimaryKey)
             {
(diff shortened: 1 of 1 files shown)


src/tools/ddl2cpp.cpp
```diff
@@ -328,15 +328,19 @@ class CxxModelPrinter
         {
             std::string type = MakeType(column, _config.forceUnicodeTextColumns);
             const auto memberName = FormatName(column.name, _config.formatType);
+
+            if (column.isForeignKey)
```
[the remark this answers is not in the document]
Yaraslaut: indeed

src/tools/ddl2cpp.cpp
```diff
@@ -328,15 +331,19 @@ class CxxModelPrinter
...
+            {
+                std::println("Column is both a foreign key and a primary key: {}.{}", table.name, column.name);
```
christianparpart, line 337: How often is this being printed for our work database?

Yaraslaut: maybe around 100 or so

src/tools/ddl2cpp.cpp
```diff
@@ -346,7 +353,19 @@ class CxxModelPrinter
 
         for (auto const& foreignKey: table.foreignKeys)
         {
-            const auto memberName = FormatName(std::string_view { foreignKey.foreignKey.column }, _config.formatType);
+
+            const auto avoidPrimaryKeyNameCollision = [&](std::string name)
+            {
+                // if  foreignKey.foreignKey.column inside renameForeignKeys
+                // we append "_fk" to the name
```
christianparpart, line 360: I think that comment is wrong, we don't append "_fk" but we prepend a configurable prefix to it, right?

src/tools/ddl2cpp.cpp
```diff
@@ -463,6 +482,7 @@ struct Configuration
     std::string schema;
     std::string modelNamespace;
     std::string outputDirectory;
+    std::string foreignKeyCollisionPrefix {};
```
christianparpart, line 485: This sounds like we definitely want to have a default here, like "fk_" or so, isn't it? (Otherwise we'll construct a C++ syntax error in the resulting C++ header if not configured explicitly).

Yaraslaut: In this case we do not have a prefix, i think this is alright, then user can check what he wants to do, instead of silently introducing some behavior

src/tools/ddl2cpp.cpp
```diff
@@ -346,7 +353,19 @@ class CxxModelPrinter
 
         for (auto const& foreignKey: table.foreignKeys)
         {
-            const auto memberName = FormatName(std::string_view { foreignKey.foreignKey.column }, _config.formatType);
+
+            const auto avoidPrimaryKeyNameCollision = [&](std::string name)
```
christianparpart, line 357:
```suggestion
            auto const avoidPrimaryKeyNameCollision = [&](std::string name)
```

src/tools/ddl2cpp.cpp
christianparpart, line 360:
```suggestion
                // we prepend `foreignKeyCollisionPrefix` to the name
```

src/tools/ddl2cpp.cpp
```diff
@@ -346,7 +353,19 @@ class CxxModelPrinter
...
+            {
+                // if  foreignKey.foreignKey.column inside renameForeignKeys
+                // we append "_fk" to the name
+                if (std::ranges::find(renameForeignKeys, foreignKey.foreignKey.column) != renameForeignKeys.end())
+                {
+                    name = std::format("{}{}", _config.foreignKeyCollisionPrefix ,name);
```
christianparpart, line 363:
```suggestion
                    name = std::format("{}{}", _config.foreignKeyCollisionPrefix, name);
```

src/tools/ddl2cpp.cpp
```diff
@@ -346,7 +353,19 @@ class CxxModelPrinter
...
+                }
+                return name;
+            };
+
+            const auto memberName = avoidPrimaryKeyNameCollision(FormatName(std::string_view { foreignKey.foreignKey.column }, _config.formatType));
```
christianparpart, line 368:
```suggestion
            auto const memberName = avoidPrimaryKeyNameCollision(FormatName(std::string_view { foreignKey.foreignKey.column }, _config.formatType));
```

PR #261, opened 2025-05-01
Load database from file inside action and test ddl2cpp header generation

Yaraslaut:


src/examples/test_chinook/header.hpp
```diff
@@ -0,0 +1,120 @@
+#include <Lightweight/DataMapper/DataMapper.hpp>
+struct Employee final
+{
+    static constexpr std::string_view TableName = "Employee";
```
christianparpart, line 4: Hm... Why's this file not containing the expected formatting, as in the empty lines that we have when calling ddl2cpp as of this week? 🤔

src/tools/ddl2cpp.cpp
```diff
@@ -393,6 +392,7 @@ class CxxModelPrinter
 
         for (SqlSchema::ForeignKeyConstraint const& foreignKey: table.externalForeignKeys)
         {
+
```
christianparpart, line 395:
```suggestion
```

src/tools/ddl2cpp.cpp
```diff
@@ -657,13 +657,39 @@ void ResolveOrderAndPrintTable(CxxModelPrinter& cxxModelPrinter, std::vector<Sql
     {
         for (auto const [index, table]: std::views::enumerate(tables))
         {
+            if (numberOfForeignKeys[index] == -1)
```
christianparpart, line 660: What does `-1` mean here? How can one have -1 foreign keys? 🤔

src/examples/test_chinook/header.hpp
Yaraslaut: i remove all empty lines, just to make it easy to compare between each other

christianparpart:
Thanks. I will review ASAP this evening

.github/workflows/build.yml
```diff
@@ -237,11 +237,22 @@ jobs:
...
-      - name: "Run DDL2CPP"
-        run: ./src/tools/test.sh
-        env:
-          DDL2CPP: "/usr/local/bin/ddl2cpp"
+        run: sudo apt install -y unixodbc-dev unixodbc libsqliteodbc odbcinst uuid-dev sqlite3 uuid-dev libyaml-cpp-dev
+      - name: "Setup ms sql server"
+        id: setup
+        run: |
+          bash ./.github/prepare-test-run.sh "MS SQL Server 2022" "./src/examples/test_chinook/Chinook_Sqlite.sql"
+      - name: "Run ddl2cpp on chinoook db"
```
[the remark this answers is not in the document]
christianparpart: lol Copilot is actually right :D

christianparpart (approved):
many thanks, but at least let's fix the typo in `Chinook` :) (as reported by copilot :) )

christianparpart:
@Yaraslaut I've applied to suggestions of Copilot. I however only wonder now if src/examples/ is the right base folder for the test. Shouldn't that go into src/tests/?

Yaraslaut:
> @Yaraslaut I've applied to suggestions of Copilot. I however only wonder now if src/examples/ is the right base folder for the test. Shouldn't that go into src/tests/?

I wanted to use this database to create some examples, so i put it there to see the header file for the database

PR #264, opened 2025-05-05
SqlDateTime: Add year/month/date/hour/minute/second/nanosecond read accessors

### LINKED ISSUE #263: SqlDateTime: get read-accessors to year/month/day/hour/minute/second
> day (1-31), month(1-12), year , hour(0-23), minute(0-59), second(0-59), millisecond(0-999)

We need these accessors in their most convenient (but safe) way. We could still return types like `std::chrono::day` etc, but since we use `TIMESTAMP_STRUCT`, it should be trivial to add accessors to them.

christianparpart:
Closes #263.

### DIFF
--- a/src/Lightweight/DataBinder/SqlDateTime.hpp
+++ b/src/Lightweight/DataBinder/SqlDateTime.hpp
@@ -84,6 +84,48 @@ struct SqlDateTime
     {
     }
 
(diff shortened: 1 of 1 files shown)


Yaraslaut (approved):
lgtm, thanks 
find it weird that the year is integer though

PR #265, opened 2025-05-05
Ensure that we do not repeat names of the members and chinook test update

Yaraslaut:
* Ensure that we never repeat names of the members, by using internal prefix `c_` to the name 
* Change structure of the test_chinook example and add simple examples of data mapper usage 
*  expose foreign key collision prefix as a cli parameter


src/examples/test_chinook/main.cpp
```diff
@@ -0,0 +1,50 @@
...
+int main()
+{
+
+    SqlConnection::SetDefaultConnectionString(SqlConnectionString {
+        "DRIVER={ODBC Driver 18 for SQL "
+        "Server};SERVER=localhost;UID=SA;PWD=QWERT1.qwerty;TrustServerCertificate=yes;DATABASE=LightweightExample" });
+    auto dm = DataMapper();
+
+    // helper function to create std::string from string_view<char16_t>
+    auto toString = [](std::basic_string_view<char16_t> str) {
```
christianparpart, line 17:
```suggestion
    auto const toString = [](std::basic_string_view<char16_t> str) {
```

src/examples/test_chinook/main.cpp
```diff
@@ -0,0 +1,50 @@
...
+        std::string result;
+        result.reserve(str.size());
+        std::ranges::copy(str, std::back_inserter(result));
+        return result;
+    };
+
+    // get all employees
+    auto empoyees = dm.Query<Employee>().All();
```
christianparpart, line 25:
```suggestion
    auto const empoyees = dm.Query<Employee>().All();
```

christianparpart (approved):
Thank you very much :)

PR #271, opened 2025-05-07
ddl2cpp: Add ability to optionally create a cumulative header, that includes them all

christianparpart:
Now we can create something like an `AllEntities.hpp` that contains the includes to all the entities, that have been generated.

The goal here is, to be able to trivially use that file to test-compile it (e.g. in our company code-base, to ensure it'll all pass smoothly, even if not all entities are used just yet)


Yaraslaut (approved):
LgtmX thanks :)

PR #274, opened 2025-05-11
Chinook update example 

Yaraslaut:
* Update entities structure 
* add some usage example in the main.cpp
* add testing to the ci 

Closes https://github.com/LASTRADA-Software/Lightweight/issues/275

### DIFF
--- a/src/Lightweight/DataBinder/UnicodeConverter.cpp
+++ b/src/Lightweight/DataBinder/UnicodeConverter.cpp
@@ -1,6 +1,7 @@
 #include "UnicodeConverter.hpp"
 
 #include <codecvt>
+#include <cstdint>
 #include <locale>
 
 #if defined(_WIN32) || defined(_WIN64)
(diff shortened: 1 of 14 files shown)


christianparpart:
How's the commit 48db5616819aa4dc821636c0bce0fa9497cc9db4 landing in here? I am confused

src/Lightweight/DataBinder/SqlNumeric.hpp
```diff
@@ -215,12 +212,11 @@ struct SqlDataBinder<SqlNumeric<Precision, Scale>>
 // clang-format off
 
 
-template <std::size_t Precision, std::size_t Scale>
-struct std::formatter<SqlNumeric<Precision, Scale>>: std::formatter<double>
+template<auto Precision, auto Scale>
```
christianparpart, line 215: why did you change from `size_t` to `auto`? Did this cause any problems?

src/Lightweight/DataBinder/UnicodeConverter.cpp
```diff
@@ -97,12 +98,72 @@ std::u16string ToUtf16(std::u8string_view u8InputString)
     return u16String;
 }
 
+std::u16string Utf8ToUtf16(const std::string& utf8)
```
christianparpart, line 101: If the input string is UTF-8, we should use `std::u8string`, or (better even), use `std::u8string_view`.

This is a place where I think, depending on method overloading is okay, so I'd have this function named analogous to the others, i.e. `std::u16string ToUtf16(std::u8string_view)`

src/Lightweight/DataBinder/UnicodeConverter.cpp
```diff
@@ -112,11 +173,7 @@ std::u16string ToUtf16(std::string const& localeInputString)
     return { reinterpret_cast<char16_t const*>(wideString.data()),
              reinterpret_cast<char16_t const*>(wideString.data() + wideString.size()) };
 #else
-    std::locale sys_locale("");
-    std::wstring_convert<std::codecvt_utf8_utf16<wchar_t>> utf16conv;
-    auto const result = utf16conv.from_bytes(localeInputString);
-    return { reinterpret_cast<char16_t const*>(result.data()),
-             reinterpret_cast<char16_t const*>(result.data() + result.size()) };
+    return Utf8ToUtf16(localeInputString);
```
christianparpart, line 176:
```suggestion
    return ToUtf16(std::u8string_view(localeINputString));
```
Then this line would be something like this.

Yaraslaut: compiler does not allow to construct `std::u8string_view(localeINputString)`

christianparpart:
Thank you, @Yaraslaut. I will review ASAP. :)

christianparpart (approved):
Thanks.

PR #276, opened 2025-05-12
Fix Update function to respect SqlRealName

Yaraslaut:
In update function we need to respect naming of the columns in the structure that can originate from SqlRealName

### DIFF
--- a/src/Lightweight/DataMapper/DataMapper.hpp
+++ b/src/Lightweight/DataMapper/DataMapper.hpp
@@ -1107,13 +1107,14 @@ void DataMapper::Update(Record& record)
 
     auto query = _connection.Query(RecordTableName<Record>).Update();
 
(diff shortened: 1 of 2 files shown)


src/Lightweight/DataMapper/DataMapper.hpp
```diff
@@ -1107,14 +1107,15 @@ void DataMapper::Update(Record& record)
 
     auto query = _connection.Query(RecordTableName<Record>).Update();
 
-    Reflection::CallOnMembers(record, [&query]<typename Name, typename FieldType>(Name const& name, FieldType const& field) {
+    Reflection::CallOnMembersWithoutName(record, [&query]<size_t I, typename FieldType>(FieldType const& field) {
```
christianparpart, line 1110: TODO: Let's name it `CallOnMembersWithIndex` or `EnumerateOnMembers` to make it sound  more familiar with std::views::enumerate? 🤔

christianparpart (approved):

PR #279, opened 2025-05-16
Fix build of where clause on data mapper update call

Yaraslaut:
Fixes build of where clause, 
* We always need to include where clause for the primary key on update query
* hack around the bad deduction of FieldType


src/Lightweight/DataMapper/DataMapper.hpp
```diff
@@ -1130,8 +1131,7 @@ void DataMapper::Update(Record& record)
     Reflection::CallOnMembers(
         record, [this, &i]<typename Name, typename FieldType>(Name const& name, FieldType const& field) mutable {
             if constexpr (FieldType::IsPrimaryKey)
-                if (!field.IsModified())
-                    _stmt.BindInputParameter(i++, field.Value(), name);
+                _stmt.BindInputParameter(i++, field.Value(), name);
```
christianparpart, line 1134: why do you do this always? an UPDATE statement should only update the fields that were updated, effectively reducing the I/O between client and server.

christianparpart (approved):
brain damaged, deluxe. Thank you :)

PR #281, opened 2025-05-19
Add query single no relations function to load entry without relations

Yaraslaut:
Adds new function that can be used to avoid heavy template instantiation for some tables, main goal is to provide only one entry point for such behavior such that users are still encouraged to use laze evaluation of the relations when possible

### DIFF
--- a/src/Lightweight/DataMapper/DataMapper.hpp
+++ b/src/Lightweight/DataMapper/DataMapper.hpp
@@ -636,6 +636,16 @@ class DataMapper
     template <typename Record, typename... PrimaryKeyTypes>
     std::optional<Record> QuerySingle(PrimaryKeyTypes&&... primaryKeys);
 
+    /// @brief Queries a single record (based on primary key) from the database without auto-loading relations.
+    ///
+    /// The primary key(s) are used to identify the record to load.
+    ///
+    /// Main goal of this function is to load record without relationships to
+    /// decrease compilation time and work around some limitations of template instantiation
+    /// depth on MSVC compiler.
+    template <typename Record, typename... PrimaryKeyTypes>
+    std::optional<Record> QuerySingleWithoutRelationAutoLoading(PrimaryKeyTypes&&... primaryKeys);
+
     /// @brief Queries a single record from the database.
     ///
     /// @return A query builder for the given Record type that will also allow executing the query.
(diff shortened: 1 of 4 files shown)


src/tests/DataMapperTests.cpp
```diff
@@ -676,6 +676,24 @@ TEST_CASE_METHOD(SqlTestFixture, "BelongsTo", "[DataMapper][relations]")
...
+    dm.Create(user);
+
+    auto email1 = Email { .id = SqlGuid::Create(), .address = "john@doe.com", .user = user };
+    dm.Create(email1);
+
+    auto actualEmail1 = dm.QuerySingleNoRelations<Email>(email1.id).value();
+
+    CHECK(actualEmail1.address == email1.address);
+    REQUIRE(!actualEmail1.user.IsLoaded());
+    // actualEmail1.user->id; // will throw
```
christianparpart, line 694: You can use `CHECK_THROWS_AS()` here

src/Lightweight/DataMapper/DataMapper.hpp
```diff
@@ -1219,6 +1227,15 @@ std::optional<Record> DataMapper::QuerySingle(PrimaryKeyTypes&&... primaryKeys)
     return { std::move(resultRecord) };
 }
 
+template <typename Record, typename... PrimaryKeyTypes>
+std::optional<Record> DataMapper::QuerySingle(PrimaryKeyTypes&&... primaryKeys)
+{
+    auto record = QuerySingleNoRelations<Record>(std::forward<PrimaryKeyTypes>(primaryKeys)...);
+    if (record)
+        ConfigureRelationAutoLoading(record.value());
```
christianparpart, line 1235:
```suggestion
    if (record)
        ConfigureRelationAutoLoading(*record);
```
Since we just tested on the availability right one line above, just use deref-overload, as `.value()` would test again (and throw, in case, which won't happen here).

src/tests/DataMapperTests.cpp
Yaraslaut: i tried and failed xD

christianparpart (approved):
LGTM, I've btw fixed the CATCH_THROWS_AS() that I suggested. The `operator->` should not be marked as `noexcept` in order to work as expected :)

Issue #278, opened 2025-05-13
SqlDynamicString not got populated correctly


FelixTheC:
I faced an issue with SqlDynamicString inside of my Entity, inside of PostProcessing duing SqlGet. 
It looks for me that the caluclation of the expected size wasn't correct. Cause I also tried it with SqlAnsiString<2056> it was working.

christianparpart:
What was the size of `N` in `SqlAnsiString<N>` that was breaking? Maybe we need to repro over here at my machine in a small session? Is this possible?

PR #283, opened 2025-05-23
Fixes loading of the dynamic string

Yaraslaut:
I believe this closes https://github.com/LASTRADA-Software/Lightweight/issues/278

it seems like we need a magic number for the string size, but not something that we actually can have as a string size


christianparpart (approved):
team workx :D

PR #284, opened 2025-05-23
Add QualifierAlignment Right to clang-format

christianparpart:
in order to force type qualifiers to the right hand side of the type

I used this PR to also fix clang-format CI check.

Some directories are excluded from this check:

- /src/examples/
- /src/tools/tests/
- /src/benchmark/ (due to tables file, we might wanna have that check here partially mabe)

### DIFF
--- a/src/Lightweight/DataBinder/BasicStringBinder.hpp
+++ b/src/Lightweight/DataBinder/BasicStringBinder.hpp
@@ -163,11 +163,8 @@ struct SqlDataBinder<AnsiStringType>
                                 nullptr);
     }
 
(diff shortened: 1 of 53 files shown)


christianparpart:
> Is there any changes with this clang format adjustments ?

@Yaraslaut oh yes, definitely. and it was braking before already. the CI test just reports success, even though it should have failed. I'll address this probably after we've merged your PRs (as much as needed) and then i'll fix the CI job (to actually fail) and also adapt the formatting globally :)

christianparpart:
@Yaraslaut i'm merging such that I can help updating the other(s)

PR #288, opened 2025-05-27
Improve test stability (focusing on running on Windows while talking to MS SQL Server)

christianparpart:
partially improving running the tests on Windows against MS SQL (via ODBC driver v17, v18)

### DIFF
--- a/src/tests/DataMapperTests.cpp
+++ b/src/tests/DataMapperTests.cpp
@@ -307,11 +307,12 @@ TEST_CASE_METHOD(SqlTestFixture, "Query.All", "[DataMapper]")
     for (auto& person: expectedPersons)
         dm.Create(person);
 
-    auto const records = dm.Query<Person>().Where(FieldNameOf<&Person::is_active>, "=", true).All();
+    auto const records =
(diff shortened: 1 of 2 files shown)


src/tests/DataMapperTests.cpp
```diff
@@ -329,11 +330,14 @@ TEST_CASE_METHOD(SqlTestFixture, "Query.First n", "[DataMapper]")
     for (auto& person: expectedPersons)
         dm.Create(person);
 
-    auto const records = dm.Query<Person>().Where(FieldNameOf<&Person::age>, ">=", 30).First(2);
+    auto const records = dm.Query<Person>()
+                             .Where(FieldNameOf<&Person::age>, ">=", 30)
+                             .OrderBy(FieldNameOf<&Person::name>, SqlResultOrdering::ASCENDING)
```
christianparpart, line 335: This is the change for the remaining lines, all the `.First(n)`'s require an `.OrderBy(...)` in order to return the result set in a reliable order.

Yaraslaut (approved):
lgtm, thanks

Issue #301, opened 2025-06-24
Structured SQL Logging API


christianparpart:
Given an application that operates on many high level objects, it may be insightful to get to know what SQL queries are given within the scope of each object.

## Current State

We have some loggers already, but maybe we need to clean up a bit:

* **SqlScopedTraceLogger:** scoped ODBC low level logger (unrelated to `SqlLogger`)
* **SqlLogger:** core logging API
* **SqlNullLogger:** specialization of `SqlLogger`, not logging anything
* **SqlStandardLogger:** creates standard log messages and prints them to stdout, ignores any SQL query logging, only warnings and errors are logged
* **SqlTraceLogger:** extends `SqlStandardLogger`, also implementing query logging
* **SqlScopedTimeLogger:** measures the time spend within the current scope and logs it to the current logger

## Goals

What we need is an interface that groups the high level objects together and allows logging any of the SQL queries (or log messages) within that group as context.

NB: It makes sense to not just allow logging to stdout, but customize it.
The suggestion is to allow customizing where to write (see `SqlStandardLogger::WriteMessage`) and to allow passing a context, in form of a string tag, which could name the high level object to be processed.

```cpp
/// mock up
void runAllComplexJob(auto& jobs)
{
    for (auto& job: jobs)
    {
        runComplexJob(job);
    }
}

void runComplexJobs(auto& job)
{
    auto sqlContext = SqlScopedLoggingContext(job.title);
    job.run();
    for (auto const& message: sqlContext.Messages())
        std::println("[{}] {}", sqlContext.Name(), message);
}

class SqlScopedLoggingContext
{
  public:
    explicit SqlScopedLoggingContext(std::string name);
    ~SqlScopedLoggingContext();

    std::string const& Name();
    std::vector<std::string> const& Messages();
};
```

Yaraslaut:
## Design mockup — not implemented yet

Writing this up as a concrete mockup so the shape can be judged before anyone commits to building it.
Three of the decisions below are hard to reverse once the API ships, so they want a call first.

### What already exists today

Part of this issue has landed since it was filed:

- **Customizable output sink** — the NB ("allow customizing where to write, see `SqlStandardLogger::WriteMessage`")
  is done. `SqlLogger::MessageWriter` + `SqlLogger::SetLoggingSink()` route every message anywhere you like
  (`SqlLogger.hpp`, documented in `docs/logging.md` § *Redirecting the output*).
[...]

PR #302, opened 2025-06-26
SqlLogger: Add SetLoggingSink(...) to allow customizing the logging sink

christianparpart:
refs #301 for allowing customizing the logging sink


Yaraslaut (approved):
lgtm, thanks, also windows error is interesting one

PR #304, opened 2025-07-01
Add test for messages struct

Yaraslaut:
Add example that is close to what we are using internally

### DIFF
--- a/src/tests/DataMapper/CreateTests.cpp
+++ b/src/tests/DataMapper/CreateTests.cpp
@@ -125,4 +125,26 @@ TEST_CASE_METHOD(SqlTestFixture, "Table with multiple primary keys", "[DataMappe
     CHECK(queriedRecord == record);
 }
 
+struct MessagesStruct
+{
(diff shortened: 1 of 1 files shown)


src/tests/DataMapperTests.cpp
```diff
@@ -1714,4 +1714,25 @@ TEST_CASE_METHOD(SqlTestFixture, "TestOptionalDynamicData", "[DataMapper]")
...
+    Field<SqlDynamicWideString<4000>, SqlRealName { "Message" }> message;
+};
+
+TEST_CASE_METHOD(SqlTestFixture, "TestMessageStruct", "[DataMapper]")
+{
+    auto dm = DataMapper {};
+    dm.CreateTable<MessagesStruct>();
+
+    MessagesStruct message {
+        .timeStamp = SqlDateTime::Now(),
```
christianparpart, line 1730: while auto-assign should work (tm) for GUIDs, it might make sense to check for that either somewhere in our test cases (as non-empty), or ensure we explicitly set it here and have that tested to be equal of our input.

christianparpart:
@Yaraslaut do we actually still need this PR?

PR #309, opened 2025-07-15
Small fixes to belongsTo and ddl2cpp

Yaraslaut:
* Fixed loading of optional field in get columns call
* ddl2cpp numeric type
* version and soversion set only on non windows systems


christianparpart (approved):
details (hidden helper implementations) should be hidden in private namespace. We use `detail` namespace for that, actually (I'm referring to `is_belongs_to` concept here e.g.)

PR #315, opened 2025-07-17
Fix loading nullable fields

### LINKED ISSUE #312: Nullable fields
When loading field we need to ensure that for the case of nullable fields, instead of GetColumn we use GetColumnNullable

christianparpart:
Closes #312

### DIFF
--- a/src/Lightweight/DataMapper/DataMapper.hpp
+++ b/src/Lightweight/DataMapper/DataMapper.hpp
@@ -161,7 +161,10 @@ void GetAllColumns(SqlResultCursor& reader, Record& record)
             ++indexFromQuery;
             if constexpr (IsField<Field>)
             {
-                field.MutableValue() = reader->GetColumn<typename Field::ValueType>(indexFromQuery);
+                if constexpr (Field::IsOptional)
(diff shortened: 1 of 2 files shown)


src/Lightweight/DataMapper/DataMapper.hpp
```diff
@@ -161,11 +161,14 @@ void GetAllColumns(SqlResultCursor& reader, Record& record)
...
-                field.MutableValue() = reader->GetColumn<typename Field::ValueType>(indexFromQuery);
+                if constexpr (Field::IsOptional)
+                    field.MutableValue() = reader->GetNullableColumn<typename Field::ValueType::value_type>(indexFromQuery);
+                else
+                    field.MutableValue() = reader->GetColumn<typename Field::ValueType>(indexFromQuery);
             }
             else if constexpr (SqlGetColumnNativeType<Field>)
             {
-                if constexpr (IsOptionalBelongsTo<Field>)
+                if constexpr (Field::IsOptional)
```
Yaraslaut, line 171: We need to ensure that we use only belongs to fields here, so this change is not needed

Yaraslaut (changes requested):
Comment and I would like to have tests for it

Yaraslaut (approved):
thanks :)

PR #316, opened 2025-07-18
Fix large variable char retrieval, requiring SQL_LONGVARCHAR rather than SQL_VARCHAR for MS SQL server

christianparpart:
@Yaraslaut Seems like we had this condition already for UTF-16 and UTF-32, just not for `char`.


Yaraslaut (approved):
Yep, thanks :)

PR #320, opened 2025-07-24
ddl2cpp: Fix treating too many SQL column types as float rather than double

### LINKED ISSUE #319: ddl2cpp: Fix translation of FLOAT / FLOAT(53) to not just use `float` but e.g. `double` if feasable
that actually caused a bug for us :)

christianparpart:
Fixes #319.


christianparpart:
@Yaraslaut I've just upgraded all my Linuxes from Fedora 41 to 42. Now I cannot spawn a docker image anymore (trying to investigate). Would you mind re-running ddl2cpp to reflect changes in this PR?

christianparpart:
@Yaraslaut okay okay, we don't repro it in our `ddl2cpp` tests it seems, as we're simply using SQLite3 rather than something else (e.g. MS SQL Server). Maybe we soon adapt. Hmmm :thinking:

Yaraslaut (approved):

Yaraslaut:
Hm, maybe we want to add something to the chinook database schema to check this in the future

PR #321, opened 2025-07-28
Atempts to introduce c++26 reflection ci and some changes to compile

Yaraslaut:
Add ci job with c++26 reflection


.github/Reflection
```diff
@@ -0,0 +1,18 @@
+FROM yaraslaut/clang-p2996:latest
```
christianparpart, line 1: shell scripts should have an appropriate file ending, such as `.sh`. The name of this file should probably be reflect what this script is doing. Reflection alone as name is only telling you that it is about reflection. But what is it doing? 😄

christianparpart: ah, reference: https://docs.docker.com/build/concepts/dockerfile/

CMakeLists.txt
```diff
@@ -17,6 +17,17 @@ include(ClangTidy)
 include(PedanticCompiler)
 
 find_package(reflection-cpp 0.4.0)
+find_package(Catch2 CONFIG REQUIRED)
+find_package(yaml-cpp CONFIG REQUIRED)
+
+option(LIGHTWEIGHT_CXX26REFLECTION "Enable experimental reflection support from c++26" OFF)
+
+if(LIGHTWEIGHT_CXX26REFLECTION)
```
christianparpart, line 25:
```suggestion
if(LIGHTWEIGHT_CXX26_REFLECTION)
```
tbh, I think that would be way more readable.

src/Lightweight/DataBinder/SqlDynamicBinary.hpp
```diff
@@ -59,7 +59,7 @@ class SqlDynamicBinary final
         _base { s, e }
     {
     }
-
+#if !defined(CXX26_REFLECTION)
```
christianparpart, line 62: why is this disabled?

src/Lightweight/SqlMigration.cpp
```diff
@@ -108,8 +108,15 @@ size_t MigrationManager::ApplyPendingMigrations(ExecuteCallback const& feedbackC
 {
     auto const pendingMigrations = GetPending();
 
+#if defined(CXX26_REFLECTION)
```
christianparpart, line 111: an ifdef sounds not nice, we could do it the more c++ way using `if constexpr (__cplusplus > 202302L)`, which basically tests for beyond C++26, hence >= C++26.

However, I am really concerned why `std::views::enumerate` is not working here, I'd rather understand that instead (but keep my above proposal or later ifdefs below)

src/Lightweight/SqlQuery/Core.hpp
```diff
@@ -964,8 +964,15 @@ detail::RawSqlCondition SqlWhereClauseBuilder<Derived>::PopulateSqlSetExpression
     using namespace std::string_view_literals;
     std::ostringstream fragment;
     fragment << '(';
+#if defined(CXX26_REFLECTION)
```
christianparpart, line 967: this really feels like a downgrade, still not getting wy we need to roll back to manual enumeration here

christianparpart (commented):
half way through, lunch time first. resuming later :)

src/Lightweight/DataBinder/SqlDynamicBinary.hpp
Yaraslaut: this is something specific to libc++ that i could not find char_traits for the unsigned char

src/Lightweight/SqlMigration.cpp
Yaraslaut: libc++ lacks implementation of enumerate

PR #323, opened 2025-07-31
Handle null termination of the strings

Yaraslaut:
Adds a test for the string with null termination so that we know what is the expected behavior

### DIFF
--- a/src/tests/DataMapper/ReadTests.cpp
+++ b/src/tests/DataMapper/ReadTests.cpp
@@ -365,6 +365,38 @@ TEST_CASE_METHOD(SqlTestFixture, "Query: SELECT into SqlVariantRow", "[DataMappe
     CHECK(record[5].TryGetStringView().value() == "c");
 }
 
+TEST_CASE_METHOD(SqlTestFixture, "Strings with null", "[String]")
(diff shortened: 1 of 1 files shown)


src/Lightweight/DataBinder/SqlDynamicString.hpp
```diff
@@ -65,7 +65,7 @@ class SqlDynamicString
     /// Returns a string view of the string.
     [[nodiscard]] LIGHTWEIGHT_FORCE_INLINE constexpr std::basic_string_view<T> str() const noexcept
     {
-        return std::basic_string_view<T> { data(), size() };
```
christianparpart, line 68: I highly doubt this is correct, while working. I rather question the correctness of `_size` to include the trailing zero byte, which it shouldn't.

Yaraslaut: i think that `_size` should include zero byte if someone explicitly placed it there, std::string constructor for example for const char* ommits zero byte, and to put it inside you need to specify size in the constructor: see the test example , so i think that `_size` should include it if someone explicitly managed to create such string

PR #324, opened 2025-07-31
[ddl2cpp] Fix translation of CHAR(n) to become a SqlTrimmedFixedString<N, T> instead of variable sized fixed capacity string.

christianparpart:
As found out by accident today, our `CHAR(n)` fields are treated whitespace-right-trimmed by the SQL server, and our client side code, and thus, so should we in Lightweight.


christianparpart:
@Yaraslaut yeah. But how?

Yaraslaut:
I would say that something in datamappertests is a good enough

christianparpart:
> I would say that something in datamappertests is a good enough

@Yaraslaut meh, nope, I don't think so. I've moved the `CxxModelPrinter` out into a new static library, named `LightweightTools` to be also linked against in LightweightTests executable to test for some very basics and to show how to add tests in the future, while also covering our current usecase in here. Well, yeah, now this PR contains 99% other changes, but at least we (finally) have an actual way of unit-testing ddl2cpp :D

Yaraslaut (changes requested):
Great, thanks a lot, please remove files that we do not need anymore, we have a script for testing ddl2cpp and also sql files with the schemas and corresponding cpp files

PR #327, opened 2025-08-02
[DataMapper] Adds `QuerySingle<Record>().Count()` and `QuerySingle<Record>().Scalar<&Record::Field>()` 

### LINKED ISSUE #322: Provide good api to fire some query once
Quite often it is needed to execute a simple query, but at the moment we need to have some boilerplate to do that
```
  auto stmt = Light::SqlStatement{ };
  auto const sqlQuery = stmt.Query(TableName).Select().
    Where("Field_one", value_one).And().
    Where("Field_second", value_second).
    Count();

  stmt.Prepare(sqlQuery);
  stmt.Execute();

  if (!stmt.FetchRow())
    return false;
  int count = stmt.GetColumn<int>(1);

```
extend api to have something like

```
std::optional<int> count = Light::Query(TableName).Select().Where(..).And().Where(...).Count().Execute();
```

christianparpart:
Closes #322

### DIFF
--- a/src/tests/DataMapperTests.cpp
+++ b/src/tests/DataMapperTests.cpp
@@ -366,7 +366,7 @@ TEST_CASE_METHOD(SqlTestFixture, "Query", "[DataMapper]")
     }
 }
 
-TEST_CASE_METHOD(SqlTestFixture, "QuerySingle.Get", "[DataMapper]")
+TEST_CASE_METHOD(SqlTestFixture, "QuerySingle", "[DataMapper]")
 {
     auto dm = DataMapper {};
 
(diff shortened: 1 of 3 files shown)


src/tests/DataMapperTests.cpp
```diff
@@ -381,10 +381,24 @@ TEST_CASE_METHOD(SqlTestFixture, "QuerySingle.Get", "[DataMapper]")
...
+    SECTION("Get() with non-existing record")
+    {
+        auto const record = dm.QuerySingle<Person>().Where(FullFieldNameOf<&Person::age>, "=", -5).Get();
+        CHECK(record.has_value() == false);
+    }
+
+    SECTION("Count()")
+    {
+        auto const count = dm.QuerySingle<Person>().Where(FieldNameOf<&Person::age>, "=", 24).Count();
+        CHECK(count == 1);
```
christianparpart, line 400: @Yaraslaut This is what you wanted, right?

src/tests/DataMapperTests.cpp
```diff
@@ -381,10 +381,32 @@ TEST_CASE_METHOD(SqlTestFixture, "QuerySingle.Get", "[DataMapper]")
...
+    }
+
+    SECTION("Single<T>()")
+    {
+        auto const result =
+            dm.QuerySingle<Person>().Where(FieldNameOf<&Person::name>, "=", "Jimbo Jones").Scalar<&Person::age>();
+        CHECK(result.has_value());
+        CHECK(result.value() == 69);
+    }
```
christianparpart, line 409: This should be what you also wanted, right?
Anything else missing?

Yaraslaut: Yes, I think this is the use case pretty much, thanks :)

Yaraslaut (approved):
Lgtm, many thanks :)

PR #328, opened 2025-08-03
[tests] Split up too large DataMapperTests.cpp into multiple smaller segments

christianparpart:
For the greater good of tests maintainability.


Yaraslaut (approved):
Great, many thanks :)

PR #329, opened 2025-08-03
[clang-tidy] Fix some reports or actively ignore some others (std C++ naming related)

christianparpart:
tried to fix *some* of the reports (mostly naming related) from clang-tidy, but explictly marked some as ignored, where we intentionally want to break, to mimmick C++ standard library API for container like classes, and similar.

The goal isto take clang-tidy reports serious again, and ideally also hard-fail CI on reports in the future.

### Actively ignored naming suggestions
those that are intentionally mimmicking standard C++ API

- `begin()` / `end()`
- `push_back()`
- `str()`
- `data()`
- `format(...)` (std::format specializations)

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -3,11 +3,11 @@ Checks: >-
   -*,
   bugprone-*,
   -bugprone-branch-clone,
+  -bugprone-crtp-constructor-accessibility,
   -bugprone-exception-escape,
   -bugprone-implicit-widening-of-multiplication-result,
   -bugprone-reserved-identifier,
   -bugprone-suspicious-include,
-  -bugprone-crtp-constructor-accessibility,
   -bugprone-chained-comparison,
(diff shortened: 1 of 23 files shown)


christianparpart:
@Yaraslaut RFC early feedback on the actively ignored ones, and the way I've marked themas ignored. I hope that doxygen will not include these comments into the documentation (should no do so, but yeah....)

Yaraslaut (approved):
lgtm, thanks :)

Yaraslaut:
also if you fixed some check you can make it warning as error

christianparpart:
> also if you fixed some check you can make it warning as error

looks like we're already failing, because we have `WarningsAsErrors: "*"` in our `.clang-tidy`. And it was complaining at me in some other recent PRs already. Not sure right now when we've enabled it, but it works.

PR #336, opened 2025-08-06
Add support for joining a table with an aliased table name

### LINKED ISSUE #334: allow joining tables to self, via adding support for table aliasing for joined tables
cc @TooManyDots 

We need query API syntax for the following construct

```sql
SELECT A.some, B.some, C.some, ... FROM Table1 AS A
INNER JOIN Table1 AS B ON A.Foo = Other.Foo
INNER JOIN Table1 AS C ON B.Foo = A.something
INNER JOIN Table1 AS D ON C.Foo = A.else
```

So the idea is to allow joining into the own table for linking records together.

christianparpart:
closes #334

### DIFF
--- a/src/tests/QueryBuilderTests.cpp
+++ b/src/tests/QueryBuilderTests.cpp
@@ -500,6 +500,56 @@ TEST_CASE_METHOD(SqlTestFixture, "SqlQueryBuilder.Join", "[SqlQueryBuilder]")
                WHERE "Table_A"."foo" = 42)"));
 }
 
+TEST_CASE_METHOD(SqlTestFixture, "Join with table aliasing", "[SqlQueryBuilder]")
+{
+    SECTION("simple case")
+    {
+        checkSqlQueryBuilder(
+            [](SqlQueryBuilder& q) {
+                return q.FromTable("That")
+                    .Select()
+                    .Fields("foo", "bar")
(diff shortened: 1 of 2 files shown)


TooManyDots (approved):

src/tests/QueryBuilderTests.cpp
```diff
@@ -500,6 +500,50 @@ TEST_CASE_METHOD(SqlTestFixture, "SqlQueryBuilder.Join", "[SqlQueryBuilder]")
...
+                return q.FromTableAs("That", "A")
+                        .Select()
+                        .Fields({"foo"sv, "bar"sv}, "A")
+                        .Fields({"foo"sv, "bar"sv}, "B")
+                        .Fields({"foo"sv, "bar"sv}, "C")
+                        .Fields({"foo"sv, "bar"sv}, "D")
+                        .InnerJoin(AliasedTableName { "That", "B" }, "foo", SqlQualifiedTableColumnName { "A", "bar" })
+                        .InnerJoin(AliasedTableName { "That", "C" }, "bar", SqlQualifiedTableColumnName { "B", "com" })
+                        .InnerJoin(AliasedTableName { "That", "D" }, "com", SqlQualifiedTableColumnName { "C", "tar" })
+                        .All();
```
christianparpart, line 535: @TooManyDots I've adapted the test case to exactly match your use-case (also adapting the `.Fields(...)`-call). I hope this also serves as a good example.

PR #338, opened 2025-08-08
DataMapper: Add `Query().All<&Record::field>()` to conveniently query a single column of the table

christianparpart:
I think it might be useful to use the same syntax to query all the rows, but only a single column of it

```cpp
auto dm = DataMapper {};
auto const ages = dm.Query<Person>()
                    .OrderBy(FieldNameOf<&Person::age>, SqlResultOrdering::ASCENDING)
                    .All<&Person::age>();
```


src/Lightweight/DataMapper/QueryBuilders.hpp
```diff
@@ -228,6 +254,53 @@ class [[nodiscard]] SqlCoreDataMapperQueryBuilder: public SqlBasicSelectQueryBui
...
+                                                 this->_query.searchCondition.tableAlias,
+                                                 this->_query.searchCondition.tableJoins,
+                                                 this->_query.searchCondition.condition,
+                                                 this->_query.orderBy,
+                                                 this->_query.groupBy));
+        SqlResultCursor reader = _stmt.GetResultCursor();
+        auto const outputColumnsBound = detail::CanSafelyBindOutputColumn<value_type>(_stmt.Connection().ServerType());
+        while (true)
+        {
+            result.emplace_back();
```
Yaraslaut, line 287: `emplace_back` returns a reference, we already know the last element, no need for `result.back()`

Yaraslaut (commented):
I am not sure that this is a good addition to the api, we get the same result by using `QuerySparse`, but if we want to provide additional way of doing it, we should share the  implementation, so i imagine that we should pass something like an additional context between calls from Query to All, to reuse the same implementation when we have `.QuerySparse<&Person::age,Person>()....All()` and `.Query<Person>().....All<&Person::age>()`

christianparpart:
> I am not sure that this is a good addition to the api, we get the same result by using `QuerySparse`, but if we want to provide additional way of doing it, we should share the implementation, so i imagine that we should pass something like an additional context between calls from Query to All, to reuse the same implementation when we have `.QuerySparse<&Person::age,Person>()....All()` and `.Query<Person>().....All<&Person::age>()`

Correct, I'm actually against too many ways of the same thing. We already have stuff that I'd like to deprecate, some early-day APIs that are however already in use internally (can be adapted, so it's fixable).

However, here, I think we could actually eliminate `QuerySparse<...>` and replace that with this solution, once we allow more than a single column.
But I think we still have a difference! `QuerySparse` creates a vector of full rows, but sparsely fills it, whereas this one does not construct any record at all, but gives you a dense vector of values.
What do you think?

Yaraslaut:
> > I am not sure that this is a good addition to the api, we get the same result by using `QuerySparse`, but if we want to provide additional way of doing it, we should share the implementation, so i imagine that we should pass something like an additional context between calls from Query to All, to reuse the same implementation when we have `.QuerySparse<&Person::age,Person>()....All()` and `.Query<Person>().....All<&Person::age>()`
> 
> Correct, I'm actually against too many ways of the same thing. We already have stuff that I'd like to deprecate, some early-day APIs that are however already in use internally (can be adapted, so it's fixable).
> 
> However, here, I think we could actually eliminate `QuerySparse<...>` and replace that with this solution, once we allow more than a single column. But I think we still have a difference! `QuerySparse` creates a vector of full rows, but sparsely fills it, whereas this one does not construct any record at all, but gives you a dense vector of values. What do you think?

one of the use case of the QuerySparse is to fill multiple columns from the structure, so for the `Person` it can be something like `age` and `name`, so then you want to return a `std::vector<Person>` where only `age` and `name` are filled while rest a empty, this is not possible with the current implementation of the `All`, I think we can replace `QuerySparse` with the templated `All` function and specialize for the case when only one argument is given to return `std::vector<deduced_type` instead of `std::vector<Field>`

christianparpart:
Moving everything behind `.Query<MainRecord>().` should have the benefit of being more unified in my opinion. Only the last function in the fluent chain of calls determins what the return type will be.

Also `QuerySingle<Record>()...Get()` was already defacto deprecated by `Query<Record>()...First()`. Both return the same `std::optional<Record>` type today. So we could also remove `QuerySingle` already (it's a pity that I realized this just now, lol).
`QuerySingle<>()` provides `Count()` also, which is duplicated logic as well (see `Query<>()...Count()`). Only `QuerySingle<>()...Scalar<>()` would need to be moved into `Query<>()` instead

Yaraslaut:
see https://github.com/LASTRADA-Software/Lightweight/pull/342/commits/4fdb408b52f1b7a9c1ce664f32ffff0b40c29f75 for the vcpkg fix

PR #346, opened 2025-08-19
Rename FullFieldNameOf to QuotedFieldNameOf

### LINKED ISSUE #343: Fix naming confusion between fully qualified field naming
We have varying naming conventions for such cases:

- `struct SqlRawColumnNameView { std::string_view value; };`
- `string_view FullFieldNameOf<F>` - yields a fully qualified and quoted table and column name
- `FieldNameOf<F>` - returns just the correct name of the field F
- `FieldNameAt<I, R>`- returns just the correct name of the I'th field in R

We still think having a dedicated type to distinguish between different kinds of names is okay, but we should streamline on what naming convention to use for the public API symbols

- `FieldNameOf<F>` / `FieldNameAt<I, R>` give the bare name (unquoted)
- `QuotedFieldNameOf<F>` gives the quoted field name, including table name

So `FullFieldName...` should e renamed to `QuotedFieldName...`

christianparpart:
closes #343.

@Yaraslaut this streamlines its plural sibling `QuotedFieldNamesOf<F...>` that was already calling `FullFieldNameOf<F>` to generate the full list. This reduces the naming convention diversion, hopefully.

### DIFF
--- a/src/Lightweight/DataMapper/QueryBuilders.hpp
+++ b/src/Lightweight/DataMapper/QueryBuilders.hpp
@@ -376,7 +376,7 @@ class [[nodiscard]] SqlCoreDataMapperQueryBuilder: public SqlBasicSelectQueryBui
     {
         auto constexpr count = 1;
         _stmt.ExecuteDirect(_formatter.SelectFirst(this->_query.distinct,
-                                                   FullFieldNameOf<Field>.string_view(),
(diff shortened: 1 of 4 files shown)


christianparpart:
@Yaraslaut wow! Sorry! haha. I just wanted to enable auto-merge and didn't think of it not waiting for an approval, lol. Well yeah, this was a very sall and simple one anyways. And hopefully an obvious change. :D

PR #352, opened 2025-08-27
Add test example with semi manual handling of the loading 

Yaraslaut:
@TooManyDots  this is something you might want to use


src/tests/DataMapper/ReadTests.cpp
```diff
@@ -572,4 +573,85 @@ TEST_CASE_METHOD(SqlTestFixture, "Retrieve optional value without output-binding
...
+};
+
+struct CustomBindingC
+{
+    Field<uint64_t, PrimaryKey::ServerSideAutoIncrement> id {};
+    Field<double> value {};
+    Field<SqlAnsiString<20>> comment {};
+};
+
+TEST_CASE_METHOD(SqlTestFixture, "CustomBinding", "[DataMapper]")
```
christianparpart, line 600: maybe we want to explain why this test case exists here in the first place? I think you (two) had something in a local discussion on why this was done here, right?

src/Lightweight/DataMapper/DataMapper.hpp
```diff
@@ -793,6 +827,53 @@ std::vector<Record> DataMapper::QueryRaw(std::string_view sqlQueryString, InputP
...
+        if constexpr (I > 0) {
+            [&]<size_t... Indices>(std::index_sequence<Indices...>) {
+                ((Indices < I ? (offset += Reflection::CountMembers<std::tuple_element_t<Indices, Tuple>>) : 0), ...);
+            }(std::make_index_sequence<I>{});
+        }
+        return offset;
+    };
+
+
+    auto const process_element = [this]<typename ElementType, size_t Offset>(ElementType& element) {
```
christianparpart, line 852: naming convention clash. `process_element` should probably be `processElement` :)

src/Lightweight/DataMapper/DataMapper.hpp
```diff
@@ -183,6 +183,40 @@ class DataMapper
...
+    ///   std::string address;
+    ///   std::string city;
+    ///   std::string country;
+    /// };
+    /// void example(DataMapper& dm)
+    /// {
+    ///   auto const sqlQueryString = R"(SELECT p.*, a.* FROM "Person" p INNER JOIN "Address" a ON p.id = a.id WHERE p.city =
+    ///   Berlin AND a.country = Germany)"; auto const records = dm.Query<Person, Address>(sqlQueryString); for (auto const&
+    ///   [person, address] : records)
+    ///   {
```
christianparpart, line 210:
```suggestion
    ///   auto const sqlQueryString = R"(SELECT p.*, a.* FROM "Person" p INNER JOIN "Address" a ON p.id = a.id WHERE p.city = Berlin AND a.country = Germany)";
    ///   auto const records = dm.Query<Person, Address>(sqlQueryString);
    ///   for (auto const& [person, address] : records)
    ///   {
```
Please proper statement composition (one per line)

src/tests/DataMapper/ReadTests.cpp
Yaraslaut: Added info to the PR and changed name of the test

src/Lightweight/DataMapper/DataMapper.hpp
```diff
@@ -183,6 +183,41 @@ class DataMapper
...
+    ///   int id;
+    ///   std::string address;
+    ///   std::string city;
+    ///   std::string country;
+    /// };
+    /// void example(DataMapper& dm)
+    /// {
+    ///   auto const sqlQueryString = R"(SELECT p.*, a.* FROM "Person" p INNER JOIN "Address" a ON p.id = a.id WHERE p.city =
+    ///   Berlin AND a.country = Germany)";
+    ///   auto const records = dm.Query<Person, Address>(sqlQueryString);
```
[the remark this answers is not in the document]
christianparpart: QueryRawToTuple would probably be more correct, since we have QueryRaw already

christianparpart (approved):

Issue #361, opened 2025-09-15
Fixed string formatting for different types


Yaraslaut:
I have some changes that are needed for the formatting of fixed string

```
diff --git a/thirdparty/Lightweight/src/Lightweight/DataBinder/SqlFixedString.hpp b/thirdparty/Lightweight/src/Lightweight/DataBinder/SqlFixedString.hpp
index 68f9dc22a5..1fddddc839 100644
--- a/thirdparty/Lightweight/src/Lightweight/DataBinder/SqlFixedString.hpp
+++ b/thirdparty/Lightweight/src/Lightweight/DataBinder/SqlFixedString.hpp
@@ -430,7 +430,11 @@ struct std::formatter<Lightweight::SqlFixedString<N, T, P>>: std::formatter<std:
     auto format(value_type const& text, format_context& ctx) const -> format_context::iterator
     {
         if constexpr (std::same_as<T, wchar_t>)
-            return std::formatter<std::string>::format(ToUtf8(text.ToStringView()), ctx);
+        {
+            auto const utf8 = Lightweight::ToUtf8(text.ToStringView());
+            auto const stdstring = std::string(reinterpret_cast<const char*>(utf8.data()), utf8.size());
+            return std::formatter<std::string>::format(stdstring, ctx);
+        }
         else
             return std::formatter<std::string>::format(text.c_str(), ctx);
     }
```

PR #365, opened 2025-09-17
Fix and test strings formatting

Yaraslaut:
Closes https://github.com/LASTRADA-Software/Lightweight/issues/361


christianparpart (approved):
Thanks, @Yaraslaut

PR #371, opened 2025-10-15
Move windows.h header into cpp files

Yaraslaut:
do not include windows.h file in headers


Yaraslaut:
It does not look like this can be done,  you have to include this header on windows

Issue #362, opened 2025-09-17
Auto loading of the relationships in different API


Yaraslaut:
We have a big difference in handling different `Query` functions from `DataMapper` in respect to configuration of lazy loading of the records, while for the functions that accepts queries we build internal functions to load dependent values, this does not happen when DataMapper returns a query builder,  for example 
```
auto allHistories =
    dm.Query<AccountHistory>().Where(FullyQualifiedNameOf<&AccountHistory::account>, "=", account2.id.Value()).All();

```
this will not create auto loading of the values,  while the following will create autoloading

```
auto allHistories =
  dm.Query<AccountHistory>(dm.FromTable(RecordTableName<AccountHistory>)
                               .Select()
                               .Fields<AccountHistory>()
                               .Where(FullyQualifiedNameOf<&AccountHistory::account>, "=", account1.id.Value())
                               .All());

```

Yaraslaut:
One of the possible solutions i see for this is to allow functionality of that kind:
'''
template<typename Type>
Type load(auto pk)
{
     return dataMapperProvider().QuerySingle<Type>(pk).Get().value_or(Type{});
}
'''
and instead of using std::function for the loading of belongs to fields, call this function to populate them. 
For this we would have to provide a global datamapper and a way to configure it in some scope. If we work with only one database this should not be a problem and the only concern is to have thread_load datamappers, but if we are working with more than one database at a time we should have a way to configure connection that belongsto loader must use, so maybe belongsto fields can hold a pointer to a datamapper from datamapper pool  and use it to load the entity from the relation
this will coupled them closer together, so we need to decide if we want to go this way

PR #380, opened 2025-10-31
[DataMapper] Have a common api to load relationships

Yaraslaut:
Closes https://github.com/LASTRADA-Software/Lightweight/issues/362

### DIFF
--- a/src/Lightweight/DataMapper/DataMapper.hpp
+++ b/src/Lightweight/DataMapper/DataMapper.hpp
@@ -150,12 +150,6 @@ class DataMapper: public std::enable_shared_from_this<DataMapper>
         return _connection;
     }
 
-    /// Returns the mutable statement reference used by this data mapper.
-    [[nodiscard]] SqlStatement& Statement() noexcept
-    {
-        return _stmt;
-    }
-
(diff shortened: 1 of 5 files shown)


src/Lightweight/DataMapper/DataMapper.hpp
```diff
@@ -463,6 +469,593 @@ class DataMapper: public std::enable_shared_from_this<DataMapper>
...
+    _dm { dm },
+    _formatter { dm.Connection().QueryFormatter() },
+    _fields { std::move(fields) }
+{
+}
+
+template <typename Record, typename Derived>
+size_t SqlCoreDataMapperQueryBuilder<Record, Derived>::Count()
+{
+    _dm.Statement().ExecuteDirect(_formatter.SelectCount(this->_query.distinct,
```
christianparpart:
```suggestion
    auto stmt = SqlStatement { _dm.Connection() };
    stmt.ExecuteDirect(_formatter.SelectCount(this->_query.distinct,
```
do not expose `SqlStatement`'s internal `SqlStatement` instance, but use it like above (also for the other cases)

christianparpart (approved):
thank you, @Yaraslaut :)

PR #382, opened 2025-11-04
[DataMapper] Extend api to reduce compilation times

Yaraslaut:
It is not always a good idea to create auto loading for the entries. 
We can rely on user to decide, since it will significantly increase compilation time

### DIFF
--- a/src/Lightweight/DataMapper/QueryBuilders.hpp
+++ b/src/Lightweight/DataMapper/QueryBuilders.hpp
@@ -15,10 +15,19 @@ namespace Lightweight
 
 class DataMapper;
 
+/// Structural type for options for DataMapper queries.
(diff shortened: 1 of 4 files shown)


src/Lightweight/DataMapper/QueryBuilders.hpp
```diff
@@ -15,10 +15,16 @@ namespace Lightweight
 
 class DataMapper;
 
+struct DataMapperOptions
+{
+    // This is the default behavior since compilation times significantly increase otherwise.
```
christianparpart:
```suggestion
    /// This is the default behavior since compilation times significantly increase otherwise.
```

christianparpart (commented):
just the minors, but LGTM otherwise. Many thanks, @Yaraslaut :)

PR #388, opened 2025-11-12
[DataMapper] Fix bug when sqlguid was create during write even if already set

Yaraslaut:
Fixes bug when we where changing SqlGuid value when creating entry in the database


src/tests/DataMapper/ReadTests.cpp
```diff
@@ -27,17 +27,29 @@ TEST_CASE_METHOD(SqlTestFixture, "Query", "[DataMapper]")
 {
     auto dm = DataMapper::Create();
 
+
```
christianparpart:
```suggestion
```
no superfluous empty lines

src/tests/DataMapper/ReadTests.cpp
```diff
@@ -27,17 +27,29 @@ TEST_CASE_METHOD(SqlTestFixture, "Query", "[DataMapper]")
...
+    SqlGuid const id1 = SqlGuid::Create();
     auto expectedPersons = std::array {
-        Person { .id = SqlGuid::Create(), .name = "John Doe", .is_active = true, .age = 42 },
+        Person { .id = id1, .name = "John Doe", .is_active = true, .age = 42 },
```
christianparpart: Why did you move this out? What's the point you're testing here for?

PR #398, opened 2025-11-27
Add simple test that is using stdexec

Yaraslaut:
Add simple asyncronous test

PR #399, opened 2025-11-27
Rework DataMapper (inter-thread and BelongsTo/etc) access

### LINKED ISSUE #397: Rework sharing DataMapper state respective to BelongsTo and external use of threading models
```cpp
class DataMapperPool
{
  public:
    // Retrieves a reference to a `thread_local` DataMapper instance.
    DataMapper& Acquire();

    // Tells the pool, that this data mapper instance is not needed by a prior call to Acquire().
    void Release(DataMapper& object);

  private:
    std::list<DataMapper> _activeInstances;
    std::list<DataMapper> _idleInstances;
};

struct ScopedDataMapper
{
  public:
    DataMapper& Get();
    DataMapper& operator()();
    gsl::not_null<DataMapper> operator->();
};

using ScopedDataMapperAccessor = std::function<ScopedDataMapper()>;

// BelongsTo<> is now holding a value of ScopedDataMapperAccessor rather than a shared_ptr<DataMapper>
// then we can also lift off the requirement to force DataMapper into shared_ptr-use only.
```

christianparpart:
closes #397

### DIFF
--- a/src/Lightweight/DataMapper/DataMapper.hpp
+++ b/src/Lightweight/DataMapper/DataMapper.hpp
@@ -85,51 +85,31 @@ namespace detail
 /// // Delete the person record from the database
 /// dm.Delete(person);
 /// @endcode
-class DataMapper: public std::enable_shared_from_this<DataMapper>
+class DataMapper
 {
-    struct PrivateTag
-    {
-        explicit PrivateTag() = default;
-    };
-
   public:
+    /// Acquires a thread-local DataMapper instance that is safe for reuse within that thread.
+    LIGHTWEIGHT_API static DataMapper& AcquireThreadLocal();
+
     /// Constructs a new data mapper, using the default connection.
-    DataMapper([[maybe_unused]] PrivateTag _):
+    DataMapper():
         _connection {},
         _stmt { _connection }
     {
     }
 
     /// Constructs a new data mapper, using the given connection.
-    explicit DataMapper(SqlConnection&& connection, [[maybe_unused]] PrivateTag _):
+    explicit DataMapper(SqlConnection&& connection):
         _connection { std::move(connection) },
         _stmt { _connection }
     {
     }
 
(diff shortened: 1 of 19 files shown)


Yaraslaut (changes requested):
I think is that we also should add some tests after merge : https://github.com/LASTRADA-Software/Lightweight/pull/398

Yaraslaut:
Merged

christianparpart:
@Yaraslaut adapted

src/Lightweight/SqlMigration.cpp
```diff
@@ -41,7 +41,7 @@ struct SchemaMigration
 std::shared_ptr<DataMapper>& MigrationManager::GetDataMapper()
 {
     if (!_mapper)
-        _mapper = DataMapper::Create();
+        _mapper = std::make_shared<DataMapper>();
 
     return _mapper;
 }
```
Yaraslaut: I think this function can simple return DataMapper& that is coming from the default data mapper pool

src/Lightweight/DataMapper/DataMapper.hpp
```diff
@@ -46,6 +46,22 @@ namespace detail
     }
 } // namespace detail
 
+class DataMapper;
+
+/// thread_local DataMapper pool to allow reusing DataMapper instances across multiple calls in the same thread.
```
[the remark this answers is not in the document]
christianparpart: This is not true. You do not need to first construct it, in order to use it. There's the `GetInstance()` method. However, I've now renamed it to `static DataMapperPool& ThreadLocal()` to be more self-explanatory on what exactly you retrieve. I've also made `Acquire()` virtual, to actually already allow overriding it (even though we don't need that right now).
I've clarified the semantics in the comment aboe.

src/Lightweight/SqlMigration.cpp
christianparpart: adapted.

Issue #403, opened 2025-12-04
AutoAssign primary key


Yaraslaut:
If structure has `AutoAssign` primary key the following code does not produce expected result

```cpp
entity entry{ .pk = 10, .message = "hello"};
dm.Create(entrty);
// entry.pk is not 10
```
the work around is following
```cpp
entity entry{ .message = "hello"};
entry.pk = 10; // This will mark it as modified
dm.Create(entrty);
// entry.pk is 10
```

PR #413, opened 2026-01-07
Set modified flag to true by default and make datamapper to set it to false during loading

Yaraslaut:
Closes https://github.com/LASTRADA-Software/Lightweight/issues/403

### DIFF
--- a/src/Lightweight/DataMapper/DataMapper.hpp
+++ b/src/Lightweight/DataMapper/DataMapper.hpp
@@ -350,9 +350,17 @@ class DataMapper
     template <typename Record>
     bool IsModified(Record const& record) const noexcept;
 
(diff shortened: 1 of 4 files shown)


src/Lightweight/DataMapper/DataMapper.hpp
```diff
@@ -1550,6 +1555,9 @@ std::optional<Record> DataMapper::QuerySingleWithoutRelationAutoLoading(PrimaryK
     if (!detail::ReadSingleResult(_stmt.Connection().ServerType(), reader, *resultRecord))
         return std::nullopt;
 
+    if(resultRecord)
```
christianparpart: Here too, missing whitespace

PR #425, opened 2026-01-16
Improve SQL migration API and introduce `dbtool`

### LINKED ISSUE #426: CLI tool dbmigrate to apply SQL migrations
Add a CLI tool `dbtool` that for now only provides migration support.
This is, it applies or is rolling back DB migrations.
`dbtool` acts as a shell as it itself does not know any DB migrations. This tool is therefore general purpose.
It does get the migrations by plugging in external libaries (.so or .dll files) that contain a list of exported db migrations via command line argument.

The tool can:

- list all applied migrations
- list all pending (not applied) migrations
- list all migrations
- apply all pending migrations
- apply a specific migration by ID (timestamp)

### command line arguments

```
dbtool migrate                - applies pending migrations to the database
dbtool list-pending           - lists pending migrations
dbtool list-applied           - lists applied migrations
dbtool apply <NAME>           - applies the migration with the given name
dbtool rollback <NAME>        - rolls back the migration with the given name
```

`dbtool` looks into a standard folder location to search for existing .so / .dll files that provide the existing migrations. This folder location is scanned for one or many library files. The folder location can be customized via command line flag.

christianparpart:
took me a while to get motivated to finish it. Turns out, it was already working and I only needed very few adaptions for revival plus then some changes to make the API somewhat more well rounded and adding rollback support plus some more tests.

closes #426.

### DIFF
--- a/src/Lightweight/SqlMigration.hpp
+++ b/src/Lightweight/SqlMigration.hpp
@@ -46,11 +46,10 @@ namespace SqlMigration
         /// Get the singleton instance of the migration manager.
         ///
         /// @return Reference to the migration manager.
-        LIGHTWEIGHT_API static MigrationManager& GetInstance()
-        {
-            static MigrationManager instance;
-            return instance;
-        }
+        /// Get the singleton instance of the migration manager.
+        ///
+        /// @return Reference to the migration manager.
+        LIGHTWEIGHT_API static MigrationManager& GetInstance();
 
         /// Add a migration to the manager.
         ///
@@ -66,7 +65,7 @@ namespace SqlMigration
         ///
         /// @param timestamp Timestamp of the migration to get.
         /// @return Pointer to the migration if found, nullptr otherwise.
-        [[nodiscard]] LIGHTWEIGHT_API MigrationBase const* GetMigration(MigrationTimestamp timestamp) const;
+        [[nodiscard]] LIGHTWEIGHT_API MigrationBase const* GetMigration(MigrationTimestamp timestamp) const noexcept;
 
         /// Remove all migrations from the manager.
         ///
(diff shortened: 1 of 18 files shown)


src/tests/MigrationTests.cpp
```diff
@@ -149,7 +149,7 @@ struct Person
 struct Order
 {
     Field<int64_t, PrimaryKey::AutoAssign> id;
-    BelongsTo<&Person::id> person;
```
Yaraslaut: I think here you need to wrap `&Person::id` with the `Member` macro since for the reflection we do not use member pointers but reflection of members

src/tools/dbtool/PluginLoader.hpp
```diff
@@ -0,0 +1,43 @@
...
+#include <filesystem>
+#include <functional>
+#include <stdexcept>
+#include <string>
+#include <vector>
+
+namespace Lightweight::Tools
+{
+
+class PluginLoader
```
Yaraslaut: can you please provide documentation for this class, what it does and where we use it

src/Lightweight/SqlMigration.hpp
```diff
@@ -133,40 +195,76 @@ namespace SqlMigration
...
+/// @brief Creates a new migration.
+/// @param timestamp Timestamp of the migration.
+/// @param description Description of the migration.
+///
+/// @note The migration will be registered with the migration manager automatically.
+///
+/// @code
+/// #include <Lightweight/SqlMigration.hpp>
+///
+/// LIGHTWEIGHT_SQL_MIGRATION(2026'01'17'23'41'20, "Create table 'MyTable'")
```
Yaraslaut: I think we should use everywhere the same syntax for the timestamps, maybe lilke
`2026_01_17_23_41_20`

src/tests/MigrationTests.cpp
```diff
@@ -201,4 +201,116 @@ TEST_CASE_METHOD(SqlMigrationTestFixture, "Migration with foreign key", "[SqlMig
...
-#endif
+
+TEST_CASE_METHOD(SqlMigrationTestFixture, "Revert Migration", "[SqlMigration]")
+{
+    using namespace SqlColumnTypeDefinitions;
+
+    auto reversibleMigration = SqlMigration::Migration(
+        SqlMigration::MigrationTimestamp { 202412102213 },
+        "reversible migration",
+        [](SqlMigrationQueryBuilder& plan) { plan.CreateTable("reversible_table").PrimaryKey("id", Integer()); },
```
Yaraslaut: I think we should support syntax similar to the data mapper queries, something like 

`plan.CreateTable<NewTable>` 

 `plan.AddColumn<&NewTable::new_field>`

such that we do not need to know what is the type of new column in the migration and have it hardcoded

src/Lightweight/SqlMigration.hpp
christianparpart: that's literally just a number, and I think it's okay to leave the semantics of that number to the caller/user. It may be just incremented (literally), or a timestamp like YYYYMMDDHHMMSS or put more readable with `'` in C++. Using `_` would require more changes. We can do later if needed.

christianparpart: @Yaraslaut Actually, I just ealized that the code comment's numeric encoding won't work.

Adding support for it, won't be easy, but adding support for what you suggested would work. BUT that would enforce the schema migrations table's key to be a string rather than a number, something that I'd like to avoid. So we're back to normal numbers without underscores and hyphens. That should be okay, I know everyone else is having it like that too, so yeah. :dagger:

PR #438, opened 2026-01-27
minors all around sql backup refactors

christianparpart:
- add `--schema-only` to `dbtool backup` / `dbtool restore`
- split up SqlBackup.cpp into Backup.cpp and Restore.cpp (simply put)

### DIFF
--- a/src/Lightweight/SqlBackup/Common.cpp
+++ b/src/Lightweight/SqlBackup/Common.cpp
@@ -7,6 +7,10 @@
 #include <format>
 #include <thread>
 
+#if defined(_WIN32)
+    #include <Windows.h>
+#endif
+
 namespace Lightweight::SqlBackup::detail
 {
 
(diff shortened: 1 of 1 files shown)


src/Lightweight/SqlBackup/Common.cpp
```diff
@@ -0,0 +1,120 @@
...
+    auto const now = std::chrono::system_clock::now();
+    return std::format("{:%FT%TZ}", now);
+}
+
+std::string FormatTableName(std::string_view schema, std::string_view table)
+{
+    if (schema.empty())
+        return std::format(R"("{}")", table);
+    return std::format(R"("{}"."{}")", schema, table);
+}
```
Yaraslaut: i think this should be a global utility function, potentially combined with the utils for the data mapper, where we have qualified column names and such

src/Lightweight/SqlBackup/Restore.cpp
```diff
@@ -0,0 +1,666 @@
...
+                }
+
+                if (hasIdentity)
+                {
+                    identityTable = FormatTableName(ctx.schema, tableName);
+                    SqlStatement { workerConn }.ExecuteDirect(std::format("SET IDENTITY_INSERT {} ON", identityTable));
+                }
+            }
+
+            {
```
Yaraslaut: why this extra scope?

src/Lightweight/SqlBackup/Restore.cpp
```diff
@@ -0,0 +1,666 @@
...
+            progress.Update({ .state = Progress::State::Error,
+                              .tableName = tableName,
+                              .currentRows = 0,
+                              .totalRows = 0,
+                              .message = std::string("AddForeignKey failed: ") + e.what() });
+        }
+    }
+}
+
+// NOLINTNEXTLINE(readability-function-cognitive-complexity)
```
Yaraslaut: this function is too big, please consider splitting it

src/Lightweight/SqlBackup/Common.cpp
```diff
@@ -0,0 +1,120 @@
...
+              .message = std::format("Connection failed, retry {}/{}: {}", attempts, settings.maxRetries, error.message) });
+
+        std::this_thread::sleep_for(CalculateRetryDelay(attempts - 1, settings));
+    }
+}
+
+std::string CurrentDateTime()
+{
+    auto const now = std::chrono::system_clock::now();
+    return std::format("{:%FT%TZ}", now);
```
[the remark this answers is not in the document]
christianparpart: @Yaraslaut What's your suggestion here?

PR #441, opened 2026-02-08
Add Reddit announcement post for project launch

christianparpart:
@Yaraslaut this is seen as a draft proposal. Feel free to freely mutate as hell in this post. We might want to actually merge it here too, so we have it on our website as well.

r/cpp post title suggestions:

- "Lightweight: Zero-overhead C++23 SQL library with DataMapper ORM, migrations, and backup/restore"

### DIFF
--- a/docs/announcement-reddit.md
+++ b/docs/announcement-reddit.md
@@ -4,7 +4,7 @@
 
 ---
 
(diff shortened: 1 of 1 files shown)


Yaraslaut (approved):
thanks a lot, nicely written announcement :)

PR #442, opened 2026-02-10
Add C++ modules support

mikomikotaishi:
This PR adds support for C++ modules.

Using statements are sorted in alphabetical order so that it is easier to maintain.

### DIFF
--- a/src/examples/test_chinook/entities/Album.hpp
+++ b/src/examples/test_chinook/entities/Album.hpp
@@ -3,10 +3,15 @@
 
 #include "Artist.hpp"
 
+#if !defined(LIGHTWEIGHT_BUILD_MODULES)
+    #include <Lightweight/DataMapper/DataMapper.hpp>
+#endif
+
 struct Album final
 {
-    static constexpr string_view TableName = "Album";
-    Field<int32_t, PrimaryKey::ServerSideAutoIncrement, SqlRealName { "AlbumId" }> AlbumId;
-    Field<SqlDynamicUtf16String<160>, SqlRealName { "Title" }> Title;
-    BelongsTo<&Artist::ArtistId, SqlRealName { "ArtistId" }> ArtistId;
+    static constexpr std::string_view TableName = "Album";
+
+    Light::Field<int32_t, Light::PrimaryKey::ServerSideAutoIncrement, Light::SqlRealName { "AlbumId" }> AlbumId;
+    Light::Field<Light::SqlDynamicUtf16String<160>, Light::SqlRealName { "Title" }> Title;
+    Light::BelongsTo<&Artist::ArtistId, Light::SqlRealName { "ArtistId" }> ArtistId;
 };
--- a/src/examples/test_chinook/entities/Track.hpp
+++ b/src/examples/test_chinook/entities/Track.hpp
@@ -5,17 +5,23 @@
 #include "Genre.hpp"
 #include "Mediatype.hpp"
 
+#if !defined(LIGHTWEIGHT_BUILD_MODULES)
+#include <Lightweight/DataMapper/DataMapper.hpp>
+#endif
+
+
 struct Track final
 {
-    static constexpr string_view TableName = "Track";
+    static constexpr std::string_view TableName = "Track";
 
-    Field<int32_t, PrimaryKey::ServerSideAutoIncrement, SqlRealName { "TrackId" }> TrackId;
-    Field<SqlDynamicUtf16String<200>, SqlRealName { "Name" }> Name;
-    BelongsTo<&Album::AlbumId, SqlRealName { "AlbumId" }, SqlNullable::Null> AlbumId;
-    BelongsTo<&Mediatype::MediaTypeId, SqlRealName { "MediaTypeId" }> MediaTypeId;
-    BelongsTo<&Genre::GenreId, SqlRealName { "GenreId" }, SqlNullable::Null> GenreId;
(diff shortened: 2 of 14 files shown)


mikomikotaishi:
Sure, I can add a test for the modules.

mikomikotaishi:
I have a test now for the modules

src/examples/test_chinook/entities/Album.hpp
```diff
@@ -3,13 +3,10 @@
 
 #include "Artist.hpp"
 
-#include <Lightweight/DataMapper/DataMapper.hpp>
```
Yaraslaut: do we need to `import Lightweight;` here instead? since this should be enough to include this header to start using the struct, and order of the includes should not matter

src/examples/test_chinook/main.cpp
```diff
@@ -1,5 +1,37 @@
...
+using Lightweight::DataMapper;
+using Lightweight::Field;
+using Lightweight::PrimaryKey;
+using Lightweight::SqlConnection;
+using Lightweight::SqlConnectionString;
+using Lightweight::SqlDateTime;
+using Lightweight::SqlDynamicUtf16String;
+using Lightweight::SqlNullable;
+using Lightweight::SqlNumeric;
+using Lightweight::SqlRealName;
```
Yaraslaut: I do not think that this is a good idea to drop all qualifiers, such as namespaces

src/examples/test_chinook/main.cpp
```diff
@@ -68,133 +92,23 @@ int main()
...
-    DumpTable<Artist>(dm);
-    DumpTable<Customer>(dm);
-    DumpTable<Employee>(dm);
-    DumpTable<Genre>(dm);
-    DumpTable<Invoice>(dm);
-    DumpTable<Invoiceline>(dm);
-    DumpTable<Mediatype>(dm);
-    DumpTable<Playlist>(dm);
-    DumpTable<Playlisttrack>(dm);
-    DumpTable<Track>(dm);
```
Yaraslaut: I guess this is a mistake, since we do not want to drop anything from the example

src/examples/test_chinook/entities/Track.hpp
```diff
@@ -5,19 +5,17 @@
...
-    Light::Field<Light::SqlNumeric<10, 2>, Light::SqlRealName { "UnitPrice" }> UnitPrice;
+    Field<int32_t, PrimaryKey::ServerSideAutoIncrement, SqlRealName { "TrackId" }> TrackId;
+    Field<SqlDynamicUtf16String<200>, SqlRealName { "Name" }> Name;
+    BelongsTo<&Album::AlbumId, SqlRealName { "AlbumId" }, SqlNullable::Null> AlbumId;
+    BelongsTo<&Mediatype::MediaTypeId, SqlRealName { "MediaTypeId" }> MediaTypeId;
+    BelongsTo<&Genre::GenreId, SqlRealName { "GenreId" }, SqlNullable::Null> GenreId;
+    Field<optional<SqlDynamicUtf16String<220>>, SqlRealName { "Composer" }> Composer;
+    Field<int32_t, SqlRealName { "Milliseconds" }> Milliseconds;
+    Field<optional<int32_t>, SqlRealName { "Bytes" }> Bytes;
+    Field<SqlNumeric<10, 2>, SqlRealName { "UnitPrice" }> UnitPrice;
```
Yaraslaut: I am slightly confused 
why qualifiers removed here?

Yaraslaut (commented):
Thanks a lot for the update, nothing big, the only point that i am worried about is qualified names, such that name resolution is as simple as possible

mikomikotaishi:
Is it a problem in this example? The symbols all belong to either the library or std, so I decided to import names into scope to make code easier to read

src/examples/test_chinook/entities/Album.hpp
mikomikotaishi: I didn't do it with the headers, because I made sure the standard library would be included beforehand.

I guess you're suggesting something like
```cpp
#ifdef LIGHTWEIGHT_BUILD_MODULES
import Lightweight;
#else
#include <Lightweight/DataMapper/DataMapper.hpp>
#endif
```
but this is a bad idea. `import` statements MUST appear all at the top of a file, and if we were to do this, we would get something like
```cpp
#include "entities/Album.hpp"
#include "entities/Artist.hpp"
```
could expand into
```cpp
import Lightweight;

struct Album final { /* ... */ };

import Lightweight;

struct Artist final { /* ... */ };
```
and an `import` statement would be coming after code, which is rejected by the compiler.

src/examples/test_chinook/main.cpp
Yaraslaut: oh, did not notice, indeed the case, we can keep it like this, this is ok

src/examples/test_chinook/entities/Album.hpp
Yaraslaut: Oh interesting, you can notice that one of the ci checks is failing that checks `ddl2cpp`. This is a small tool that generates this header files for the database, see: https://github.com/LASTRADA-Software/Lightweight/actions/runs/22004219452/job/63585918320

So I guess we need to find a way to deal with this hear

Yaraslaut:
Thanks a lot @mikomikotaishi  i pushed you commits into master with some additions to the ddl2cpp, the main reason i added a guard and did not touch anything else, is because we need to ensure that we have this files as a proper valid headers, if we rely on user to include something before including this headers, this can be problematic

PR #469, opened 2026-04-20
[SqlBackup] Add tests for --schema-only backup/restore

christianparpart:
The `--schema-only` flag for `dbtool backup` / `dbtool restore` is fully wired on master (CLI → `BackupSettings::schemaOnly` / `RestoreSettings::schemaOnly` → library short-circuits), but there was no test coverage for either branch. This PR adds tests at the library and CLI levels so the behaviour is guarded against regressions.

## Changes

- `src/tests/SqlBackup/Tests.cpp`: add `SqlBackup: Schema-only Backup and Restore` — performs a schema-only backup, opens the resulting ZIP with libzip, and asserts that `metadata.json` is present, no `data/*` entries exist, and `checksums.json` is absent; then schema-only-restores and asserts the table exists with 0 rows.
- `src/tests/SqlBackup/Tests.cpp`: add `SqlBackup: Schema-only restore from full backup` — produces a normal (data-bearing) backup and then restores it with `schemaOnly = true`, confirming the data chunks in the archive are ignored.
- `src/tests/test_dbtool.py`: extend the CLI integration script with two new steps — (1) `dbtool backup --schema-only` followed by a Python-side `zipfile` inspection of the archive layout, and (2) a `--dry-run --schema-only` invocation whose stdout must mention schema-only mode.

Verified across all three DBMS backends (`--test-env=sqlite3`, `postgres`, `mssql2022`) — all 115 `[SqlBackup]` cases pass on each, and the Python integration script completes end-to-end on SQLite.

### DIFF
--- a/src/tests/SqlBackup/Tests.cpp
+++ b/src/tests/SqlBackup/Tests.cpp
@@ -1633,4 +1633,123 @@ TEST_CASE("SqlBackup: Restore rejects unsupported format version", "[SqlBackup]"
     }
 }
 
+// =============================================================================
+// Schema-only Backup / Restore Tests
+// =============================================================================
+
+TEST_CASE("SqlBackup: Schema-only Backup and Restore", "[SqlBackup]")
+{
+    ScopedFileRemoved const backupFileCleaner { BackupFile };
+
+    SetupDatabase(); // creates test_table and inserts 3 rows
+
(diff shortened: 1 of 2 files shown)


Yaraslaut (approved):
Lgtm, thanks :)

PR #470, opened 2026-04-20
Add migration dependencies, audit metadata, and SQLite column idempotency

christianparpart:
The migration system gains three capabilities that close gaps around ordering, auditability, and cross-dialect idempotency.

## Changes

- **Migration dependencies.** A migration can declare `MigrationMetadata::dependencies`; `GetPending()` returns a topologically-ordered list, and `ApplyPendingMigrations`/`ApplySingleMigration` reject unknown deps or cycles with a descriptive error. When nothing declares deps, the existing timestamp ordering is preserved unchanged.
- **Audit metadata.** `MigrationBase` exposes optional `GetAuthor()` / `GetDescription()` overrides, and `schema_migrations` now carries `author`, `description`, and `execution_duration_ms` (all nullable). `ApplySingleMigration` wall-clock-times the `Up()` run; `MarkMigrationAsApplied` records the author/description but leaves duration null.
- **SQLite column idempotency.** `AddColumnIfNotExists` / `DropColumnIfExists` used to emit a comment plus a raw `ALTER TABLE`, which would fail on a repeat apply. The SQLite formatter now emits a `LIGHTWEIGHT_SQLITE_GUARD` sentinel; the migration executor parses it, consults `pragma_table_info`, and runs the DDL only when needed. PostgreSQL and SQL Server paths are unchanged.


src/Lightweight/SqlMigration.cpp
```diff
@@ -122,11 +130,334 @@ MigrationManager::MigrationList MigrationManager::GetPending() const noexcept
...
+        std::string columnName;
+    };
+
+    /// Parse the leading sentinel (if any) from a SQL script.
+    ///
+    /// @param script The SQL script possibly starting with a `-- LIGHTWEIGHT_SQLITE_GUARD: ...` line.
+    /// @return The parsed guard and position after the sentinel line, or std::nullopt if absent.
+    [[nodiscard]] std::optional<std::pair<SqliteGuard, size_t>> TryParseSqliteGuard(std::string_view script)
+    {
+        constexpr auto marker = std::string_view { "-- LIGHTWEIGHT_SQLITE_GUARD:" };
```
Yaraslaut, line 343: we should reuse this in the SQLiteFormatter or use here the one defined in the formatter, to avoid any issues

src/Lightweight/SqlMigration.cpp
```diff
@@ -122,11 +130,334 @@ MigrationManager::MigrationList MigrationManager::GetPending() const noexcept
...
+        }
+
+        if (shouldExecute)
+            (void) stmt.ExecuteDirect(body);
+    }
+} // namespace
+
 void MigrationManager::ApplySingleMigration(MigrationBase const& migration)
 {
+    // Check declared dependencies: every dep must already be applied.
```
Yaraslaut:
```suggestion
    // Check declared dependencies: every dependency must already be applied.
```

src/Lightweight/SqlMigration.hpp
```diff
@@ -49,11 +49,11 @@ namespace SqlMigration
...
-        MigrationTimestamp timestamp;   ///< The timestamp of the verified migration.
-        std::string_view title;        ///< The title of the verified migration.
-        std::string storedChecksum;    ///< The checksum stored in the database.
-        std::string computedChecksum;  ///< The checksum computed from the current migration definition.
-        bool matches;                  ///< Whether the stored and computed checksums match.
+        MigrationTimestamp timestamp; ///< The timestamp of the verified migration.
+        std::string_view title;       ///< The title of the verified migration.
+        std::string storedChecksum;   ///< The checksum stored in the database.
+        std::string computedChecksum; ///< The checksum computed from the current migration definition.
+        bool matches;                 ///< Whether the stored and computed checksums match.
```
Yaraslaut, line 56: do not comment inline, add a comment on top

src/tests/MigrationTests.cpp
```diff
@@ -656,6 +656,300 @@ TEST_CASE_METHOD(SqlMigrationTestFixture, "CreateTableIfNotExists", "[SqlMigrati
...
+    // Despite declaration order, apply should respect the dependency.
+    auto migrationA = SqlMigration::Migration(
+        SqlMigration::MigrationTimestamp { 202601010001 }, "dep base", [](SqlMigrationQueryBuilder& plan) {
+            plan.CreateTable("dep_a").PrimaryKey("id", Integer());
+        });
+
+    auto migrationB = SqlMigration::Migration(
+        SqlMigration::MigrationTimestamp { 202601010002 },
+        "dep dependent",
+        SqlMigration::MigrationMetadata { .dependencies = { SqlMigration::MigrationTimestamp { 202601010001 } } },
```
Yaraslaut: Hm, i would prefer to have a less bugprone syntax here, for instance 
```
MigrationMedatada { .dependencies = { TimeStamp<decltype(migrationA)> }}
```
or 
```
MigrationMedatada { .dependencies = { decltype(migrationA)::TimeStamp }}
```

Yaraslaut (commented):
Some comments, also i would like to see an example and tests with real usage, where we would use macro and static variable to create a list of migrations

src/Lightweight/SqlMigration.hpp
christianparpart: I don't mind either way. What's the problem with this one? It's a well known syntax for it? hmmm.... :thinking:

Yaraslaut (approved):

PR #471, opened 2026-04-21
Add release tagging for SQL migrations

christianparpart:
Long-running projects accumulate hundreds of migration timestamps over years of development, but nothing in the library connects them to the release versions users actually talk about. This PR lets migrations be grouped under a version string: declare `LIGHTWEIGHT_SQL_RELEASE("6.7.0", 20240501120000)` alongside the migrations that shipped with that cut, and dbtool surfaces the mapping. Releases live purely in code — they compile into migration plugins and are imported by dbtool the same way migrations are, so no `schema_releases` table is needed.

## Changes

- **Release API** on `MigrationManager`: `RegisterRelease`, `RemoveAllReleases`, `GetAllReleases`, `FindReleaseByVersion`, `FindReleaseForTimestamp`, `GetMigrationsForRelease`. A migration `M` belongs to release `R` iff `prev_release_ts < M.timestamp <= R.highestTimestamp`.
- **`LIGHTWEIGHT_SQL_RELEASE(version, highestTimestamp)`** macro registers a release via a RAII `ReleaseRegistrar` at static-init time. Duplicate versions and duplicate timestamps are rejected.
- **dbtool**: new read-only `releases` command prints each release with its `applied` / `pending` / `partial` / `empty` status; new write-command `rollback-to-release <VERSION>` resolves the version to its timestamp and reverts everything newer. `CollectMigrations` imports release declarations from plugin `MigrationManager` instances alongside migrations.
- **`LIGHTWEIGHT_SQL_MIGRATION_REVERSIBLE` / `LIGHTWEIGHT_SQL_MIGRATION_DOWN`** macros fill a gap in the original `LIGHTWEIGHT_SQL_MIGRATION`, which hardcoded an empty `Down()` and `HasDownImplementation() == false`. The two hand-rolled migrations in the dummy plugins are converted to the new macros.
- Unit tests cover registration, duplicate detection, version/timestamp lookup, range-bounded migration retrieval, and rollback-to-release semantics. `test_dbtool.py` exercises `releases` and `rollback-to-release` end-to-end.

### DIFF
--- a/src/Lightweight/SqlMigration.hpp
+++ b/src/Lightweight/SqlMigration.hpp
@@ -78,6 +78,25 @@ namespace SqlMigration
         size_t totalRegistered {};     ///< Total number of registered migrations
     };
 
+    /// Associates a software release with the highest migration timestamp present at release time.
+    ///
+    /// Releases are declared in source (typically a migration plugin) via
+    /// `LIGHTWEIGHT_SQL_RELEASE(version, highestTimestamp)`. They let tools answer questions like
+    /// "which migrations belong to release 6.7.0?" or "roll back everything applied after 6.7.0".
+    ///
+    /// The `highestTimestamp` is an inclusive upper bound: a migration `M` belongs to the release
+    /// iff `prev_release_ts < M.timestamp <= highestTimestamp`, where `prev_release_ts` is the
+    /// previous release's timestamp (or 0 if there is none).
+    ///
+    /// @ingroup SqlMigration
+    struct MigrationRelease
+    {
+        /// Human-readable version string, e.g. "6.7.0".
+        std::string version;
+        /// Highest migration timestamp contained in this release (inclusive).
+        MigrationTimestamp highestTimestamp;
+    };
+
     /// Main API to use for managing SQL migrations
     ///
     /// This class is a singleton and can be accessed using the GetInstance() method.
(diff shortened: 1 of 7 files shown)


src/Lightweight/SqlMigration.hpp
```diff
@@ -496,11 +556,25 @@ namespace SqlMigration
...
+    ///
+    /// @ingroup SqlMigration
+    struct ReleaseRegistrar
+    {
+        /// Registers `{version, highestTimestamp}` with the singleton manager.
+        ReleaseRegistrar(std::string version, MigrationTimestamp highestTimestamp)
+        {
+            MigrationManager::GetInstance().RegisterRelease(std::move(version), highestTimestamp);
+        }
+    };
```
Yaraslaut: if this is not for public to use, put inside `detail` namespace

src/Lightweight/SqlMigration.hpp
```diff
@@ -543,3 +617,87 @@ namespace SqlMigration
...
+/// @brief Companion to @c LIGHTWEIGHT_SQL_MIGRATION_REVERSIBLE defining the out-of-line `Down()` body.
+///
+/// Must be used in the same translation unit and follow the matching
+/// `LIGHTWEIGHT_SQL_MIGRATION_REVERSIBLE(timestamp, ...)` declaration.
+///
+/// @param timestamp Timestamp of the migration whose Down() body follows.
+///
+/// @ingroup SqlMigration
+#define LIGHTWEIGHT_SQL_MIGRATION_DOWN(timestamp) \
+    void Migration_##timestamp::Down(Lightweight::SqlMigrationQueryBuilder& plan) const
```
Yaraslaut: I think this is not needed, if you want some migration to be reversible, you should put it inside migration registration part, not somewhere else in the code, there should be only one source of information about the migration

PR #474, opened 2026-04-23
Qt 6 GUI for SQL migrations on top of dbtool

christianparpart:
The headline change is **dbtool-gui**, a Qt 6 / QML front-end for browsing, previewing, and applying SQL migrations against any configured connection profile. The CLI / library work in the rest of this PR — new `dbtool` subcommands, profile + secrets handling, ODBC enumeration, cross-backend migration semantics — grew up while building the GUI and is needed to support it. To keep the public Lightweight API clean, the helpers that only `dbtool` and `dbtool-gui` use live in a new tools-private static library (`src/tools/shared/`, target `tools_shared`); they are *not* installed and *not* part of the `Lightweight::Lightweight` ABI.

The commits on this branch are deliberately split: the foundational CLI / library work first, the GUI on top.

## Headline change — dbtool GUI

`src/tools/dbtool-gui/` (opt-in via `LIGHTWEIGHT_BUILD_GUI=ON`): Qt 6 / QML front-end for `dbtool` against any configured connection profile.

- Migration plan view grouped by `LIGHTWEIGHT_SQL_RELEASE` markers, with per-migration SQL preview and one-click apply / revert / backup-restore actions (`MigrationRunner`, `BackupRunner` running on `QThread` workers).
- Connection panel backed by `Config::ProfileStore` + `Secrets::SecretResolver`, with ODBC DSN auto-discovery and an optional `qtkeychain`-based credential backend.
- Ad-hoc SQL query panel (`SqlQueryRunner`, `SqlSyntaxHighlighter`, `SqlResultModel`) so the GUI is not migration-only.
- Light/dark `ThemeController` and a small reusable QML component set (`Card`, `StatusPill`, `KineticListView`, `WheelScrollAmplifier`, ...).

## Tools-private shared library (`src/tools/shared/`, target `tools_shared`)

These helpers are consumed by `dbtool`, `dbtool-gui`, `lup2dbtool`, and the test suite. They are STATIC, not installed, and not part of the public Lightweight CMake export. yaml-cpp is linked PRIVATE on `tools_shared`, so it no longer propagates to the installed Lightweight package.

- **`Config::ProfileStore`** — YAML-backed store for named connection profiles, used by both `dbtool` and the GUI so credentials no longer live in plaintext on disk.
- **`Secrets::SecretResolver`** with `EnvBackend` / `FileBackend` / `StdinBackend` — pluggable indirection for credentials referenced from profiles. `FileBackend` refuses files with mode wider than `0600`.
- **`Odbc::DataSourceEnumerator`** — wraps `SQLDataSources` and driver enumeration so the GUI can populate a DSN dropdown.
- **`CodeGen::SplitFileWriter`** — codegen helper used by `lup2dbtool` to emit one C++ file per migration with a shared CMake snippet.

## Lightweight library changes (public API)

- **Structured `MigrationException`** — carries operation / timestamp / title / step index / failed SQL / driver message, so both the CLI and GUI can show actionable error context without parsing `what()`.
- **Bounded migration apply/preview** — `MigrationManager::ApplyPendingMigrationsUpTo` / `PreviewPendingMigrationsUpTo` / `FindReleaseByVersion`, the forward counterparts to the existing rollback path. Plus `FoldRegisteredMigrations` as a primitive for callers that want to collapse a chain.
- **Cross-backend migration semantics for legacy SQL corpora** — composite `WhereExpression` / `SetExpression` on `Update` / `Delete`, idempotent variants of `AddColumn` / `DropColumn` / `AddForeignKey` / `DropIndex` across SQLite / PostgreSQL / MSSQL, deterministic FK constraint names, a SQLite `ALTER TABLE` rebuild path for `AddForeignKey` / `DropForeignKey`, and per-migration compat policy with a `lup-truncate` renderer.
- **`HardReset`** — drops tables in batched transactions; preserved-tables list compares by unqualified name only, so an unqualified plan resolves correctly against engine-specific default schemas (`dbo` / `public` / none).
- **MSSQL UTF-8 string literals** — `NCHAR` concatenation so non-BMP characters survive round-trip.
- **`SqlSchema` cross-engine introspection fixes** — `SQLForeignKeys` row grouping now keys on `FK_NAME` (with a SQLite `PRAGMA` fallback), plus assorted reader gaps that previously caused identical migrations to look like drift across engines.
- **Robustness fixes** — `SqlConnection` now surfaces the original driver diagnostic when `Connect()` fails; `SqlStatement` tolerates `SQL_NO_DATA` from `SQLExecDirect`; `SqlConnectInfo::EnsureSqliteDatabaseFileExists` bootstraps a missing file-based SQLite database.

## `dbtool` — new commands and reworks

- Migrated onto `Config::ProfileStore` + `Secrets::SecretResolver` (drops the direct `yaml-cpp` link from `dbtool` and the bespoke `~/.config/dbtool/dbtool.yml` parsing); adds `--profile <name>` and surfaces the latest applied release in `status`. An explicit `--connection-string` without `--profile` no longer silently merges profile defaults (`schema`, `pluginsDir`, ...) — the typed connection string pins the backend.
- `dbtool exec <QUERY>` — ad-hoc SQL execution against a profile.
- `dbtool hard-reset` — drop all tables in batched transactions (used by the GUI and CI).
- `dbtool unicode-upgrade-tables` — bulk UTF-8 upgrade for legacy MSSQL schemas.
- `dbtool rewrite-checksums` — rewrite stored migration checksums after an authorised content change, with plugin policy propagation.
- `dbtool migrate-to-release <VERSION>` — forward-direction counterpart of `rollback-to-release`; resolves the version to its declared `highestTimestamp` and applies pending migrations up to that point.
- `dbtool status` — aligned label column and reports the latest available release alongside the latest applied one.
- `--show-examples` — long-form examples moved out of the default `--help`.
- `dbtool` SqlLogger now routes through the standard logger.

## `lup2dbtool` and `LupMigrationsPlugin`

- New `LupSqlParser`, `WhereClauseParser`, `StringUtils`, expanded `SqlStatementParser` and `CodeGenerator` lift the legacy `init_m_*.sql` / `upd_m_*.sql` corpus into one C++ file per migration (`lup_{version}.cpp`), with `--emit-cmake` for the shared CMake snippet.
- Per-file encoding detection (UTF-8 / UTF-16 / Windows-1252) with strict explicit-mode flags for reproducible CI runs.
- `--force-unicode` is now the default (with `--no-force-unicode` opt-out).
- `LupMigrationsPlugin` installs the `lup-truncate` compat policy by timestamp (cutoff extended to all LUP migrations), scrubs proprietary paths from placeholder comments, and cold-start bootstraps `lup2dbtool` itself when invoked from a parent build that is still mid-configure (with `CONFIGURE_DEPENDS` glob refresh).

## Build / packaging

- New `tools_shared` STATIC library at `src/tools/shared/` — see above. Linked PUBLIC by `dbtool_lib` (so `dbtool-gui` inherits it transitively) and by `lup2dbtool_lib`; tests inherit it via those.
- `LIGHTWEIGHT_BUILD_GUI` is gated by a Qt 6 auto-probe in the top-level CMake; if Qt isn't found the option is silently downgraded to OFF instead of failing the configure.
- `Lightweight_CPMAddPackage` wrapper around `CPMAddPackage` so clang-tidy no longer fires on `_deps/` (stdexec, libzip, nlohmann_json, reflection-cpp, tracy). First-party tidy coverage unchanged (89 invocations, all under `src/`).
- MSVC C4251/C4275 suppression for shared builds.
- Installed Lightweight header tree no longer contains `Config/`, `Secrets/`, `Odbc/`, or `CodeGen/`; yaml-cpp is no longer in the installed `Lightweight-targets` exports.

## Documentation

- `docs/dbtool.md` — covers the new subcommands and profile flow.

## Risk

- Surface area: large, but the GUI is opt-in (`LIGHTWEIGHT_BUILD_GUI=ON`) and the new `dbtool` subcommands are additive.
- Behaviour change: `dbtool` no longer reads `~/.config/dbtool/dbtool.yml` directly — existing users must migrate to a profile. `--connection-string` alone no longer pulls profile defaults; users who relied on that need to also pass `--profile`.
- Public API: no breaking changes to the installed `Lightweight::Lightweight` headers — but the helpers that previously sat in `src/Lightweight/{Config,Secrets,Odbc,CodeGen}/` (which were never exported by the C++20 module and never consumed inside the core library) have moved to `src/tools/shared/`. They remain in the `Lightweight::` namespace; only the `#include` path and link target change for tool-side consumers.

## Coverage

- New unit tests: `ConfigProfileStoreTests`, `SecretResolverTests`, `DataSourceEnumeratorTests`, plus large additions to `Lup2DbtoolTests`, `MigrationTests`, and `QueryBuilderTests`.
- `MigrationTests` now verifies `HardReset` / `UnicodeUpgradeTables` post-conditions through `SqlSchema::ReadAllTables` instead of `sqlite_schema`, so the cases run on every backend the suite is parameterised over.

## Test plan

- [x] Build with `LIGHTWEIGHT_BUILD_GUI=ON`, launch `dbtool-gui`, exercise apply / revert / backup-restore against a SQLite profile.
- [x] Confirm the SQL preview, ad-hoc query panel, and theme toggle work end-to-end in the GUI.
- [x] Build with `LIGHTWEIGHT_BUILD_GUI=OFF` (default) and confirm CLI-only build is unchanged.
- [x] Run `dbtool exec`, `dbtool hard-reset`, `dbtool unicode-upgrade-tables`, `dbtool rewrite-checksums`, `dbtool migrate-to-release` against a throwaway profile.
- [x] Run the full unit test suite under each enabled backend (`sqlite3`, `mssql2022`, `postgres`).
- [x] Verify `cmake --install` does not ship `Config/`, `Secrets/`, `Odbc/`, or `CodeGen/` under `include/Lightweight/`, and that the installed `Lightweight-targets.cmake` no longer references `yaml-cpp`.


src/tools/dbtool-gui/Models/ReleaseListModel.hpp
```diff
@@ -0,0 +1,62 @@
...
+#include <vector>
+
+#include <QtCore/QAbstractListModel>
+#include <QtQmlIntegration/QtQmlIntegration>
+
+namespace Lightweight
+{
+namespace SqlMigration
+{
+    class MigrationManager;
```
Yaraslaut, line 24: please include directly the header

src/tools/dbtool-gui/Models/MigrationListModel.hpp
```diff
@@ -0,0 +1,127 @@
...
+#include <QtCore/QAbstractListModel>
+#include <QtQmlIntegration/QtQmlIntegration>
+
+// Forward declaration: see MigrationRunner.hpp for why the full include is
+// kept out of moc-parsed headers.
+namespace Lightweight
+{
+namespace SqlMigration
+{
+    class MigrationManager;
```
Yaraslaut, line 30: include the header

src/tools/dbtool-gui/ThemeController.hpp
```diff
@@ -0,0 +1,91 @@
...
+
+#pragma once
+
+#include <QtCore/QObject>
+#include <QtQml/QQmlEngine>
+
+namespace DbtoolGui
+{
+
+class ThemeController: public QObject
```
Yaraslaut, line 25: tbh I am now sure about the theme switching, especially if we need to maintain it

src/tools/dbtool-gui/SqlQueryRunner.hpp
```diff
@@ -0,0 +1,76 @@
...
+    void finished(int rowCount, qint64 elapsedMs, QString status);
+    /// SQL or driver error. `sqlState` may be empty for non-SQL errors;
+    /// `nativeError` is 0 in that case as well.
+    void errorOccurred(QString message, QString sqlState, int nativeError);
+
+  private:
+    void SetBusy(bool busy);
+
+    QThreadPool _pool;
+    std::atomic<bool> _busy { false };
```
Yaraslaut, line 80: add a `std::stop_source` member to stop execution if possible

src/tools/dbtool-gui/QtKeychainBackend.hpp
Yaraslaut, line 1: why do we need this file since ?

src/tools/dbtool-gui/BackupRunner.cpp
```diff
@@ -0,0 +1,175 @@
...
+      private:
+        QObject* _target;
+    };
+
+} // namespace
+
+BackupRunner::BackupRunner(QObject* parent):
+    QObject(parent)
+{
+    _pool.setMaxThreadCount(1);
```
Yaraslaut, line 102: why only one thread? i would like to get as many as possible to make it faster

Yaraslaut (changes requested):
left some comments, i think that we need to have parallel execution of the backup/restore to make it fast

src/tools/dbtool-gui/ThemeController.hpp
christianparpart: so your suggestion is? drop it? hard-code it? I think we do not need to maintain anything like this. We either have it or we don't. If it turns out to not be useful in the future (or it breaks), so be it. It can be dropped (if we want) or latently fixed whenever time permits :)

src/tools/dbtool-gui/BackupRunner.cpp
christianparpart: that's some time ago, but I remember that it wasn't doing good in terms of performance improvements.

src/tools/dbtool-gui/Models/ReleaseListModel.hpp
christianparpart: Tried it — the comment turns out to be correct. Replacing the forward-decl with `#include <Lightweight/SqlMigration.hpp>` makes `qmltyperegistrar` re-anchor `DbtoolGui` under `Lightweight::`: the generated `dbtool-gui_qmltyperegistrations.cpp` then emits `qmlRegisterTypesAndRevisions<Lightweight::DbtoolGui::AppController>(...)` and clang-cl errors with `'DbtoolGui' is not a class, namespace, or enumeration`. Keeping the forward-decl + comment.

src/tools/dbtool-gui/Models/MigrationListModel.hpp
christianparpart: Same finding as in the `ReleaseListModel.hpp` thread — including the header reproduces the `qmltyperegistrar` namespace re-anchor and the build fails. The forward-decl workaround stays.

src/tools/dbtool-gui/BackupRunner.cpp
christianparpart: Reconsidered after your PR #484 thread-pool comment. `SqlBackup::Backup`/`Restore` already spawn `concurrency` worker threads internally (`SqlBackup.cpp:471`/`:1056`); the GUI was just passing 1 and never exercising it. Now passing `clamp(hardware_concurrency(), 1, 8)`. `setMaxThreadCount(1)` on the Qt pool stays — it serialises GUI-initiated backup/restore so the progress pane + cancel button remain coherent; the library is where the parallelism lives. MS SQL is internally clamped to 1 by the library (`SqlBackup.cpp:342`/`:1034`), so no GUI-side per-DBMS check needed.

src/tools/dbtool-gui/ThemeController.hpp
christianparpart: Going to leave it as-is for now. If it bit-rots on a Qt/KDE update, we drop it then — no active maintenance commitment.

Yaraslaut (approved):

PR #479, opened 2026-04-28
Fix PostgreSQL Unicode round-trip on Windows by switching ODBC to W variants

christianparpart:
The Lightweight ODBC layer used the A (ANSI) variants of every ODBC entry point — `SQLDriverConnectA`, `SQLConnectA`, `SQLPrepareA`, `SQLExecDirectA`, the bare `SQLTables`/`SQLColumns`/`SQLForeignKeys`/`SQLPrimaryKeys`/`SQLStatistics` (which resolve to A by default), `SQLGetInfoA`, `SQLGetDiagRecA`, and friends. psqlODBC inspects which variant of `SQLDriverConnect` was called *first* on a DBC handle and pins the entire handle into either ANSI or Unicode mode for its lifetime. ANSI mode runs every `SQL_C_CHAR` payload through the system codepage — cp1252 on Windows — so UTF-8 bytes ≥ 0x80 round-trip as their cp1252-decoded form re-encoded back to UTF-8. Concretely, on Windows + PostgreSQL:
- `u8\"Hellö\"` round-tripped as `HellÃ¶`
- `u\"Unicode \U0001F601\"` round-tripped as `Unicode ??` (surrogate pair lost)
- `100× ä` (200 UTF-8 bytes) into `VARCHAR(100)` was rejected with `value too long for type character varying(100)` because the driver re-encoded into 400 bytes for length checks
- BatchManager + corner-case tests with embedded `\n` round-tripped as `\r\n` (a separate psqlODBC default — `LFConversion=1`)
- `dbtool status` against PostgreSQL emitted nothing despite running normally and exiting 0 (libpq.dll's unload bypassed the C++ runtime's stream flush)

Linux runs were happy because there the cp1252 step is a no-op for the typical UTF-8 locale, so CI never caught any of this.

Calling the W variants from connect onward puts the handle into Unicode-app mode and lets the driver itself do encoding conversion correctly, including reassembling UTF-16 surrogate pairs into the server's UTF-8 wire format. After that, the existing `BasicStringBinder` PostgreSQL hatches that pre-converted to UTF-8 and bound as `SQL_C_CHAR` become wrong (they put us back on the cp1252 path), so this PR removes them and routes every Unicode payload through `SQL_C_WCHAR` uniformly.

The fix is committed in independently-shippable steps so each one is green on its own.

## Changes

- `SqlError`: collect ODBC diagnostics via `SQLGetDiagRecW`, decode the SQLWCHAR buffers back to UTF-8 with a noexcept-safe fallback. The body moves from the header into `SqlError.cpp` so the new UnicodeConverter dependency stays out of every translation unit that includes `SqlError.hpp`.
- `SqlConnection`: every connect / attr / info entry point switched to its W form. A new file-local `Utf8ToUtf16(string_view)` wrapper does the cast from UTF-8 `std::string` to `std::u16string`; a `GetInfoStringW` helper centralizes the `SQLGetInfoW` byte-vs-character convention. A `static_assert` pins `sizeof(SQLWCHAR) == sizeof(char16_t)` so the `reinterpret_cast` to `SQLWCHAR*` remains sound on Windows (`wchar_t`) and Linux unixODBC (`unsigned short`) alike.
- `SqlStatement`: `SQLPrepareW` + `SQLExecDirectW`. The query (UTF-8 `std::string_view`) is converted to a local `std::u16string` per call.
- `SqlSchema`: explicit `SQLTablesW` / `SQLColumnsW` / `SQLForeignKeysW` / `SQLPrimaryKeysW` / `SQLStatisticsW` with a small `OdbcWideArg` struct that holds the converted buffer and exposes the `(pointer, length)` pair the W introspection calls expect, with correct `nullptr` / zero semantics for empty arguments.
- `SqlVariant`: `SQLColAttributeW` (pure rename — string buffer is null).
- `SqlScopedTraceLogger` + test `Utils.hpp`: trace logger and test fixture moved to W. Also fixes a latent use-after-free on the trace-file path.
- `BasicStringBinder`: drop the PostgreSQL UTF-8 hatches from the `std::u16string` / `std::u32string` / `std::u8string` parameter binders; add a PostgreSQL-only UTF-16 / `SQL_C_WCHAR` path on the `std::string` binder that treats input as UTF-8 by convention.
- `scripts/tests/.test-env.yml`: add `LFConversion=0` to the postgres connection strings — independent psqlODBC default that translates LF↔CRLF on Windows and breaks byte-exact `\n` round-trips.
- `tests/DataBinderTests`: new `Unicode round-trip across binders` test that pins `Hellö`, U+1F601 emoji, and mixed BMP+supplementary-plane round-trips through every backend.
- `tests/Utils`: cherry-pick of the upstream `[tests] Utils: suppress Windows blocking error dialogs` change so abort-style failures don't block unattended runs.
- `dbtool`: switch stdout to unbuffered so `std::println` output isn't lost when libpq.dll's atexit path bypasses the C++ runtime's stream flush. `tests/test_dbtool.py` now decodes subprocess output as UTF-8 explicitly so dbtool's progress glyphs don't crash the reader thread on Windows.

Final state on Windows: SQLite (5065 assertions), MSSQL 2022 (5070 assertions), and PostgreSQL (5072 assertions) all pass cleanly across the full LightweightTest suite plus the dbtool integration test.

### DIFF
--- a/src/Lightweight/DataBinder/BasicStringBinder.hpp
+++ b/src/Lightweight/DataBinder/BasicStringBinder.hpp
@@ -83,7 +83,8 @@ namespace detail
             result->resize(result->size() * 2);
             auto* const bufferStart = result->data() + writeIndex;
             size_t const bufferCharsAvailable = result->size() - writeIndex;
-            sqlResult = SQLGetData(stmt, column, CType, bufferStart, static_cast<SQLLEN>(bufferCharsAvailable), indicator);
+            sqlResult = SQLGetData(
+                stmt, column, CType, bufferStart, static_cast<SQLLEN>(bufferCharsAvailable * sizeof(CharType)), indicator);
         }
         return sqlResult;
     }
--- a/src/Lightweight/DataBinder/BasicStringBinder.hpp
+++ b/src/Lightweight/DataBinder/BasicStringBinder.hpp
@@ -145,6 +145,23 @@ namespace detail
                     (CharType const*) u32String.data() + u32String.size(),
                 };
             }
+
+            // The UTF-16 binder applies StringTraits::PostProcessOutputColumn (if present) so that
+            // FIXED_SIZE_RIGHT_TRIMMED strings strip the trailing CHAR(N)/NCHAR(N) padding the server
+            // returned. The UTF-32 / UTF-8 path constructs *result above, which preserves the padding,
+            // so we have to invoke the same post-process here. The synthetic indicator is the
+            // result's own byte count: PostProcessOutputColumn divides by sizeof(CharType) to get
+            // the char count, and TrimRight then walks back over trailing whitespace/null.
+            using StringTraits = SqlBasicStringOperations<StringType>;
+            if constexpr (requires { StringTraits::PostProcessOutputColumn(result, *indicator); })
+            {
+                if (*indicator != SQL_NULL_DATA && *indicator != SQL_NO_TOTAL)
+                {
+                    auto const syntheticIndicator =
+                        static_cast<SQLLEN>(StringTraits::Size(result) * sizeof(CharType));
+                    StringTraits::PostProcessOutputColumn(result, syntheticIndicator);
+                }
+            }
         });
 
         return SQLBindCol(stmt,
@@ -416,15 +433,9 @@ struct SqlDataBinder<Utf16StringType>
                                     Utf16StringType const& value,
                                     SqlDataBinderCallback& cb) noexcept
     {
-        // Bind UTF-16 input directly via SQL_C_WCHAR for every backend, including
-        // PostgreSQL. The previous PG-specific hatch — pre-converting to UTF-8 and
-        // binding as SQL_C_CHAR / SQL_VARCHAR — was a workaround for psqlODBC running
-        // in ANSI mode on Windows, where SQL_C_CHAR is treated as the system codepage
-        // (cp1252) and re-encoded as UTF-8. With SqlConnection now opening every handle
-        // via SQLDriverConnectW the driver is in Unicode app mode, where SQL_C_WCHAR
-        // input is converted to the server's encoding by the driver itself — including
-        // correct handling of UTF-16 surrogate pairs for supplementary-plane code points
-        // (emoji, CJK extension, etc.).
+        // Bind UTF-16 input directly via SQL_C_WCHAR for every backend; the driver
+        // converts to the server's encoding, including correct handling of UTF-16
+        // surrogate pairs for supplementary-plane code points (emoji, CJK extension B+).
         using CharType = StringTraits::CharType;
         auto const* data = StringTraits::Data(&value);
         auto const sizeInBytes = StringTraits::Size(&value) * sizeof(CharType);
@@ -526,10 +537,7 @@ struct SqlDataBinder<Utf32StringType>
                                     SqlDataBinderCallback& cb) noexcept
     {
         // Always go via UTF-16 + SQL_C_WCHAR; the driver handles encoding conversion
-        // for the target server. See the matching comment on the std::u16string binder
-        // above — same rationale, the previous PostgreSQL-specific UTF-8 hatch was
-        // a workaround for ANSI-mode psqlODBC and is no longer needed now that the
-        // connection is opened via SQLDriverConnectW.
+        // for the target server.
         auto u16String =
             std::make_shared<std::u16string>(ToUtf16(detail::SqlViewHelper<Utf32StringType>::View(value)));
         cb.PlanPostExecuteCallback([u16String = u16String]() {}); // Keep the string alive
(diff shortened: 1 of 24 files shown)


src/tests/DataBinderTests.cpp
```diff
@@ -1324,4 +1324,67 @@ TEST_CASE_METHOD(SqlTestFixture, "SqlWideString read from VARCHAR column", "[Sql
...
+// (`Hellö` came back as `HellÃ¶`) and dropped UTF-16 surrogate pairs (emoji came back
+// as `??`). The fix is the entire `[Lightweight] {SqlConnection,SqlStatement,
+// SqlSchema} ... via W variants` series of commits plus the BasicStringBinder cleanup;
+// these tests run on every backend so a future regression on any one of them surfaces.
+TEST_CASE_METHOD(SqlTestFixture, "Unicode round-trip across binders", "[SqlDataBinder][Unicode]")
+{
+    auto stmt = SqlStatement {};
+    stmt.MigrateDirect([](SqlMigrationQueryBuilder& migration) {
+        migration.CreateTable("UnicodeRoundTrip").Column("value", SqlColumnTypeDefinitions::NVarchar { 64 });
+    });
```
Yaraslaut, line 1339: i would like to add some tests and use something like `dm.CreateTable` together with all possible string types, we already have a test for non unicode types like that, so maybe we should extend it there, also, we should check how trimming works for this types

src/Lightweight/SqlOdbcWide.hpp
```diff
@@ -0,0 +1,69 @@
...
+// ODBC W variants assume `SQLWCHAR` is layout-compatible with `char16_t` (true on
+// Windows where it's `wchar_t`, and on Linux unixODBC where it's `unsigned short` —
+// both 16-bit). Anything else and the reinterpret_casts below would silently corrupt.
+static_assert(sizeof(SQLWCHAR) == sizeof(char16_t),
+              "ODBC W variants require a 16-bit code unit; SQLWCHAR shape mismatch");
+
+/// @brief Converts a UTF-8 string view into a `std::u16string` for ODBC W variants.
+/// Distinct from `ToUtf16(std::string const&)` in `UnicodeConverter.hpp`, which on
+/// Windows interprets its input as the system ANSI codepage (CP_ACP) — the very
+/// trap this whole switch to W variants exists to side-step.
```
Yaraslaut: the comment about the switch to the W variants is out of context if not specified further

src/Lightweight/DataBinder/SqlVariant.cpp
```diff
@@ -22,8 +22,15 @@ SQLRETURN SqlDataBinder<SqlVariant>::GetColumn(
     SQLHSTMT stmt, SQLUSMALLINT column, SqlVariant* result, SQLLEN* indicator, SqlDataBinderCallback const& cb) noexcept
 {
     SQLLEN columnType {};
+    // Use the W variant — the call passes nullptr for the string buffer (we only fetch
+    // the numeric SQL_DESC_TYPE), so it's a near-pure rename, but it keeps the entire
+    // ODBC call surface on the Unicode-app path the rest of the library now uses.
```
Yaraslaut:
```suggestion
    // ODBC call surface on the Unicode-app path the rest of the library is using.
```
I prefer comments about the current state, not about the past state

src/Lightweight/DataBinder/SqlVariant.cpp
```diff
@@ -22,8 +22,15 @@ SQLRETURN SqlDataBinder<SqlVariant>::GetColumn(
...
+    // Note: SQLColAttributeA takes a *signed* `SQLSMALLINT ColumnNumber` (legacy ODBC 2.0
+    // signature), while SQLColAttributeW takes the modern *unsigned* `SQLUSMALLINT`. The
+    // pre-existing `static_cast<SQLSMALLINT>(column)` is therefore wrong (and a
+    // clang-tidy sign-conversion error) when paired with the W variant — drop it.
```
Yaraslaut: not sure that this is helpful since it just comments about the signature of the new function

src/Lightweight/DataBinder/BasicStringBinder.hpp
```diff
@@ -608,47 +601,36 @@ struct SqlDataBinder<Utf8StringType>
...
-            }
-        }
-        std::unreachable();
+        // Always go via UTF-16 + SQL_C_WCHAR; the driver handles encoding conversion
+        // for the target server. Same rationale as the std::u16string / std::u32string
+        // binders above: the previous PostgreSQL-specific UTF-8 hatch (bind UTF-8 bytes
+        // as SQL_C_CHAR / SQL_VARCHAR) only worked when the connection happened to be in
+        // ANSI mode on a UTF-8 system locale (Linux). On Windows it ran the bytes through
+        // cp1252 → UTF-8 a second time, doubling each byte ≥ 0x80 — emoji and `ö`
+        // round-tripped wrong, and 200-byte UTF-8 payloads tripped VARCHAR-length checks.
```
Yaraslaut: again, while the comment describes the current changes, once we do not see the previous state, it does not give us any info, since now we do not have PostgreSQL specific handling

src/Lightweight/DataBinder/BasicStringBinder.hpp
```diff
@@ -508,57 +524,34 @@ struct SqlDataBinder<Utf32StringType>
...
-                                        static_cast<SQLLEN>(sizeInBytes),
-                                        indicator);
-            }
-        }
-        std::unreachable();
+        // Always go via UTF-16 + SQL_C_WCHAR; the driver handles encoding conversion
+        // for the target server. See the matching comment on the std::u16string binder
+        // above — same rationale, the previous PostgreSQL-specific UTF-8 hatch was
+        // a workaround for ANSI-mode psqlODBC and is no longer needed now that the
+        // connection is opened via SQLDriverConnectW.
```
Yaraslaut: see another comment

src/Lightweight/DataBinder/BasicStringBinder.hpp
```diff
@@ -380,54 +415,35 @@ struct SqlDataBinder<Utf16StringType>
...
-        std::unreachable();
+        // Bind UTF-16 input directly via SQL_C_WCHAR for every backend, including
+        // PostgreSQL. The previous PG-specific hatch — pre-converting to UTF-8 and
+        // binding as SQL_C_CHAR / SQL_VARCHAR — was a workaround for psqlODBC running
+        // in ANSI mode on Windows, where SQL_C_CHAR is treated as the system codepage
+        // (cp1252) and re-encoded as UTF-8. With SqlConnection now opening every handle
+        // via SQLDriverConnectW the driver is in Unicode app mode, where SQL_C_WCHAR
+        // input is converted to the server's encoding by the driver itself — including
+        // correct handling of UTF-16 surrogate pairs for supplementary-plane code points
+        // (emoji, CJK extension, etc.).
```
Yaraslaut: see other comment

Yaraslaut (approved):
Great that we managed to make api more uniform, just few small comments about test coverage and comments

PR #481, opened 2026-04-30
Add dbtool fold: emit self-contained baseline from registered migrations

christianparpart:
Adds a new offline `dbtool fold` subcommand that walks all registered migrations and emits a single self-contained baseline — either a `.cpp` migration plugin or a `.sql` script — that reproduces the post-migration schema and `schema_migrations` rows from an empty database. Useful for collapsing a long migration history into a fast-to-apply starting point, or for shipping a snapshot baseline alongside a release.

The command is **purely offline**: it never opens a DB connection, never queries a live schema. It loads plugins, walks each migration's `Up()` plan in timestamp order, folds the cumulative effect into a per-table view + chronological data steps, and emits via the existing `ToSql()` formatter path so each dialect's CREATE TABLE / INSERT codegen stays the single source of truth.

## Changes

- New `MigrationManager::FoldRegisteredMigrations(formatter, upToInclusive)` primitive — pure plan-walk, returns `PlanFoldingResult` (per-table state, creation order, indexes, chronological data steps, in-range releases). Used by the new module and available to any future caller.
- New `Lightweight::MigrationFold` library module under `src/Lightweight/MigrationFold/`:
  - `Folder` — thin facade plus `ResolveUpTo()` which accepts an empty string (latest registered release), a numeric timestamp, or a release version string.
  - `SqlEmitter` — emits a flat dialect-specific `.sql` script, including a `CREATE TABLE schema_migrations` and a stamping `INSERT` per folded timestamp so a freshly-loaded DB looks identical to a real apply-all run.
  - `CppEmitter` — emits a `.cpp` baseline plugin wrapped in `LIGHTWEIGHT_SQL_MIGRATION`, with optional `--emit-cmake` and `--max-lines-per-file` for splitting very large baselines across multiple files.
- Shared `CodeGen/SplitFileWriter` helper that bin-packs blocks within a per-file line budget; used by `CppEmitter` and intentionally factored out for reuse.
- `dbtool fold --output FILE [--up-to X] [--dialect D] [--emit-cmake] [--plugin-name N] [--max-lines-per-file N]` — output format is picked from the file extension. `.sql` requires `--dialect` (sqlite, postgres, mssql, mysql); `.cpp` is dialect-agnostic. Dispatched **before** `SetupConnectionString` since fold never touches a DB; uses a connection-less `GetMigrationManagerOffline` variant.
- Unit tests: 10 fold cases (create + altercolumn, drop-table cleanup, chronological ordering, `--up-to` truncation, RawSql passthrough, column rename FK propagation, release-range filtering, `ResolveUpTo` parsing) + 4 `SplitFileWriter` cases + 2 emitter round-trip cases. Green against sqlite3, mssql2022, and postgres.


src/Lightweight/MigrationFold/SqlEmitter.cpp
```diff
@@ -0,0 +1,169 @@
...
+        }
+    }
+} // namespace
+
+void EmitSqlBaseline(FoldResult const& fold, SqlEmitOptions const& options)
+{
+    if (!options.formatter)
+        throw std::runtime_error("EmitSqlBaseline: formatter is required");
+
+    std::ofstream out(options.outputPath);
```
Yaraslaut: can we use std::string and flush it only once in a file, when everything is done, i am not the biggest fan of `<<` and streams. Also this will remove the need for `WriteSchemaMigrationsSeed` and other functions to have first argument

src/Lightweight/MigrationFold/CppEmitter.hpp
```diff
@@ -0,0 +1,50 @@
...
+
+/// @brief Configuration for `EmitCppBaseline`.
+struct CppEmitOptions
+{
+    /// Output `.cpp` file. May be a `<stem>.cpp` — the writer will split into
+    /// `<stem>_part01.cpp`, `<stem>_part02.cpp`, ... when needed.
+    std::filesystem::path outputPath;
+    /// Threshold for splitting the body across multiple `.cpp` files. Zero
+    /// disables splitting and emits a single file.
+    std::size_t maxLinesPerFile = 5000;
```
Yaraslaut, line 24: Hm, i guess i am missing something, our migrations are declerative, and if we are folding migrations, then we have only one point of declaration, how can we split this into multiple files?

src/Lightweight/MigrationFold/CppEmitter.cpp
```diff
@@ -0,0 +1,377 @@
...
+                else if constexpr (std::is_same_v<T, Time>)
+                    return std::string(kPrefix) + "Time {}";
+                else if constexpr (std::is_same_v<T, Timestamp>)
+                    return std::string(kPrefix) + "Timestamp {}";
+                else if constexpr (std::is_same_v<T, Char>)
+                    return std::format("{}Char {{ {} }}", kPrefix, t.size);
+                else if constexpr (std::is_same_v<T, NChar>)
+                    return std::format("{}NChar {{ {} }}", kPrefix, t.size);
+                else if constexpr (std::is_same_v<T, Varchar>)
+                    return std::format("{}Varchar {{ {} }}", kPrefix, t.size);
```
Yaraslaut: i think here it is easier to use `overloaded` with the visitor, not one lambda

src/Lightweight/MigrationFold/CppEmitter.cpp
```diff
@@ -0,0 +1,377 @@
...
+    // For now use the simple single-file emitter. The shared `SplitFileWriter` is
+    // wired up below for the multi-file path, but we'd need a part-coordinator
+    // emitter analogous to lup2dbtool's `WriteSplitMainFile` to use it productively.
+    // Instead we measure the produced text up front and split only when it exceeds
+    // `maxLinesPerFile` lines, by writing the same body verbatim into N files where
+    // each file is its own self-registering migration sourced from a different
+    // timestamp slot. (For the LUP plugin the typical baseline body is well under
+    // any reasonable threshold, so this rarely fires.)
+    auto const body = BuildSingleFileBody(fold);
+    std::ofstream out(options.outputPath);
```
Yaraslaut, line 368: here as well, I think it is cleaner to create strings and then flush everything at once

Yaraslaut (approved):
Thanks, left few small comments, mostly nitpicks

src/Lightweight/MigrationFold/CppEmitter.hpp
christianparpart: because that file may become quite big (imagine having 500+ tables with many of them having really a lot of columns), then the one migration TU can be too big for a single .cpp file to be compiled on your machine. In my case, it did let clang-tidy OOM-kill my 64GB laptop :)
The approach then is to split the single migration across multiple functions that are invoked from the single folded migration.

Yaraslaut:
closing now, if we will need it, then we should recover this RP

PR #482, opened 2026-04-30
Run Windows CI suite against all supported databases

christianparpart:
The Linux CI matrix runs the full `LightweightTest` suite against SQLite3, MS SQL Server, and PostgreSQL, but Windows CI has only ever exercised MS SQL Server (LocalDB). ODBC driver and Unicode behaviour differs non-trivially between Windows and Linux drivers — the psqlODBC `LFConversion` quirk, MSSQL Driver 17 vs 18, and the SQLite3 driver registration name are all Windows-only concerns — so a Linux-only multi-DB matrix is not enough to catch Windows regressions. This PR brings the Windows job to parity.

The structure mirrors the existing Linux `ubuntu_build_cc_matrix` → `dbms_test_matrix` split.

## Changes

- `windows:` no longer runs the suite inline. After the `windows-cl-debug` build it stages the install tree (binaries, test plugins, `test_dbtool.py`) plus vcpkg-deployed runtime DLLs and uploads it as the `windows-cl-debug-tests` artifact. `windows-clangcl-debug` remains build-only.
- New `windows_dbms_test_matrix` job downloads that artifact and runs the full suite plus `test_dbtool.py` in three parallel legs: `sqlite3`, `mssql` (LocalDB), and `postgres`. Each leg installs only the driver/server it needs, waits for readiness, and writes a single `.test-env.yml` selected via `--test-env=`.
- MSSQL stays on LocalDB on Windows; cross-version (2017/2019/2022) coverage is already provided by the Linux matrix. Tests run only on `windows-cl-debug` to keep Windows runner cost in check.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -150,25 +150,112 @@ jobs:
           VCPKG_ROOT: "${{ runner.workspace }}/vcpkg"
       - name: "Build"
         run: cmake --build --preset ${{ matrix.preset }}
-      - name: Setup SQL Server
-        if: ${{ endsWith(matrix.preset, 'windows-cl-debug') }}
-        shell: powershell
+      # NB: Don't run tests here; the windows_dbms_test_matrix job below runs
+      # the suite against each supported database in parallel.
+      - name: "Install (stage tests for upload)"
+        if: ${{ matrix.preset == 'windows-cl-debug' }}
+        shell: pwsh
+        run: |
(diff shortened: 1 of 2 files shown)


Yaraslaut:
i think we should also try to use docker images to connect to, not only local db, not sure that they are guaranteed to be the same

Yaraslaut (approved):
thanks a lot :)

PR #484, opened 2026-04-30
Add dbtool diff command for comparing two databases

christianparpart:
Operators reviewing migration drift across environments today have no first-class way to ask "do these two databases match?" — the only options are diffing schema dumps by hand or running a full migration replay and hoping nothing falls out. This PR introduces `dbtool diff <SOURCE-A> <SOURCE-B>`, a read-only command that reports schema and data differences between two ODBC-reachable databases (profile names or raw `DRIVER=…` connection strings) and exits with a clean status code suitable for CI gating.

The schema diff pairs tables and columns by name only, so it produces a clean (zero-difference) result when comparing the same migration run across SQLite, PostgreSQL, and MSSQL — engine-specific schema labels (`dbo` vs `public` vs `""`), driver-reported size noise, and TINYINT/SMALLINT promotion are folded into a logical projection rather than treated as drift. The data diff streams `SELECT * FROM t ORDER BY <pk>` on both sides and reports added / removed / changed rows, with per-table error containment so one un-diffable table doesn't abort the whole run.

To make the command self-contained the PR also lands the connection-profile and secrets infrastructure dbtool needs: a YAML-backed `Lightweight::Config::ProfileStore`, a chained `Lightweight::Secrets::SecretResolver` (env / file / stdin backends), and a `MigrationException` that carries structured error fields instead of an opaque message string.

## Changes

- New `Lightweight::SqlSchema::DiffSchemas` and `DiffTableData` (`src/Lightweight/SqlSchemaDiff.{hpp,cpp}`, `SqlDataDiff.{hpp,cpp}`) — pure-logic comparators with cross-engine equivalence baked in.
- Cross-engine introspection fixes in `SqlSchema.cpp`: FK row grouping keyed on `FK_NAME`, MSSQL non-clustered index path via `sys.indexes`, SQLite `DECIMAL(p,s)` recovery from the dialect string, `SQL_VARBINARY` mapped to `VarBinary` (variable-width).
- New `dbtool diff` subcommand with `--schema-only`, `--max-rows N`, `--no-color`. Exit codes: `0` = equivalent, `1` = differs, `2` = error.
- `DiffRenderer` (`src/tools/dbtool/DiffRenderer.{hpp,cpp}`) renders schema + data diffs as a colored bordered markdown table to stdout. ANSI auto-disables on non-tty / `--no-color`.
- Vendored `src/Lightweight/tui/` (markdown table renderer, SGR builder, grapheme-aware Unicode helpers) + libunicode 0.9.0 via `cmake/FindOrFetchLibunicode.cmake`.
- `Lightweight::Config::ProfileStore` — multi-profile YAML at `~/.config/dbtool/dbtool.yml`, with profile→connection-string flattening and a shared `ProfileToConnectionString` helper.
- `Lightweight::Secrets::SecretResolver` — chained backends (`env:`, `file:`, `stdin:`) with explicit and bare-reference resolution.
- Structured `MigrationException` carrying `operation`, `timestamp`, `title`, `failedStepIndex`, `failedSql`, `sqlState`, `nativeErrorCode`. dbtool prints these as a multi-line report.
- dbtool migrates off its bespoke yaml-cpp loader onto `ProfileStore`/`SecretResolver`; logging routes through `StandardLogger`.
- 21 Catch2 cases for the diff library (16 `[SqlSchemaDiff]`, 5 `[SqlDataDiff]`); 14 cases for `[ProfileStore]` / `[SecretResolver]`.
- Doc: `docs/dbtool.md` gains a `diff` section; `docs/lup-legacy-migration-plan.md` is added as a related design note.

Stacks below `feature/migrations-gui`, which will be rebased on top of this once it lands.

### DIFF
--- a/src/Lightweight/CMakeLists.txt
+++ b/src/Lightweight/CMakeLists.txt
@@ -99,6 +99,12 @@ set(HEADER_FILES
     SqlServerType.hpp
     SqlStatement.hpp
     TracyProfiler.hpp
+
+    tui/MarkdownInline.hpp
+    tui/MarkdownTable.hpp
+    tui/SgrBuilder.hpp
+    tui/Style.hpp
+    tui/Unicode.hpp
 )
 
 set(SOURCE_FILES
@@ -175,14 +181,45 @@ target_sources(LightweightTools PRIVATE
 )
 target_link_libraries(LightweightTools PUBLIC Lightweight)
 
+# LightweightTui — vendored terminal/markdown rendering helpers (tui/).
+# Kept as a separate static library so its symbols don't need to cross the
+# Lightweight DLL ABI on Windows. libunicode stays a private detail.
+add_library(LightweightTui STATIC
+    tui/MarkdownInline.hpp
+    tui/MarkdownTable.cpp
+    tui/MarkdownTable.hpp
+    tui/SgrBuilder.cpp
+    tui/SgrBuilder.hpp
+    tui/Style.hpp
+    tui/Unicode.cpp
+    tui/Unicode.hpp
+)
+add_library(Lightweight::Tui ALIAS LightweightTui)
+target_compile_features(LightweightTui PUBLIC cxx_std_23)
+target_include_directories(LightweightTui PUBLIC
+    $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/..>
+)
+if(NOT WIN32)
+    set_target_properties(LightweightTui PROPERTIES POSITION_INDEPENDENT_CODE ON)
+endif()
+
+include("${CMAKE_SOURCE_DIR}/cmake/FindOrFetchLibunicode.cmake")
+if(TARGET unicode::unicode)
+    target_link_libraries(LightweightTui PRIVATE unicode::unicode)
+elseif(TARGET unicode::core)
+    target_link_libraries(LightweightTui PRIVATE unicode::core)
+endif()
+
 if(CLANG_TIDY_EXE)
     set_target_properties(Lightweight PROPERTIES CXX_CLANG_TIDY "${CLANG_TIDY_EXE}")
     set_target_properties(LightweightTools PROPERTIES CXX_CLANG_TIDY "${CLANG_TIDY_EXE}")
+    set_target_properties(LightweightTui PROPERTIES CXX_CLANG_TIDY "${CLANG_TIDY_EXE}")
 endif()
 
 # Enable coverage instrumentation for library targets
 enable_coverage_for_target(Lightweight)
 enable_coverage_for_target(LightweightTools)
+enable_coverage_for_target(LightweightTui)
 
 target_include_directories(Lightweight PUBLIC
     $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/..>
--- a/src/Lightweight/CMakeLists.txt
+++ b/src/Lightweight/CMakeLists.txt
@@ -93,7 +93,9 @@ set(HEADER_FILES
     SqlMigration.hpp
     SqlOdbcWide.hpp
     SqlQueryFormatter.hpp
+    SqlDataDiff.hpp
     SqlSchema.hpp
+    SqlSchemaDiff.hpp
     SqlScopedLock.hpp
     SqlScopedTraceLogger.hpp
     SqlServerType.hpp
@@ -137,7 +139,9 @@ set(SOURCE_FILES
 
     SqlQueryFormatter.cpp
     SqlScopedLock.cpp
+    SqlDataDiff.cpp
     SqlSchema.cpp
+    SqlSchemaDiff.cpp
     SqlStatement.cpp
     SqlTransaction.cpp
     Utils.cpp
--- a/src/Lightweight/SqlDataDiff.cpp
+++ b/src/Lightweight/SqlDataDiff.cpp
@@ -1,0 +1,303 @@
+// SPDX-License-Identifier: Apache-2.0
+
+#include "SqlDataDiff.hpp"
+
+#include "DataBinder/SqlVariant.hpp"
+#include "SqlConnection.hpp"
+#include "SqlServerType.hpp"
+#include "SqlStatement.hpp"
+
+#include <chrono>
+#include <format>
+#include <ranges>
+#include <string>
+
+namespace Lightweight::SqlSchema
+{
+
+namespace
+{
+
+    /// Returns the dialect-appropriate delimiter pair for identifiers.
+    [[nodiscard]] std::pair<char, char> IdentifierDelimiters(SqlServerType server) noexcept
+    {
+        switch (server)
+        {
+            case SqlServerType::MICROSOFT_SQL:
+                return { '[', ']' };
+            case SqlServerType::MYSQL:
+                return { '`', '`' };
+            case SqlServerType::POSTGRESQL:
+            case SqlServerType::SQLITE:
+            case SqlServerType::UNKNOWN:
+                return { '"', '"' };
+        }
+        return { '"', '"' };
+    }
+
+    /// Quotes an identifier and escapes embedded delimiter chars by doubling them
+    /// (the standard SQL convention; works for all dialects we target).
+    [[nodiscard]] std::string Quote(std::string_view identifier, SqlServerType server)
+    {
+        auto const [open, close] = IdentifierDelimiters(server);
+        auto out = std::string {};
+        out.reserve(identifier.size() + 2);
+        out += open;
+        for (auto const c: identifier)
+        {
+            out += c;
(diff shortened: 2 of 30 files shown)


docs/lup-legacy-migration-plan.md
Yaraslaut, line 1: this file should not be in the PR and it also gives some internal info

src/tools/dbtool/main.cpp
```diff
@@ -262,45 +289,49 @@ struct Options
...
     }
-
-    return "dbtool.yml";
-#endif
+    return SqlConnectionString {};
 }
 
-void LoadConfig(Options& options)
+/// Loads a profile from a `ProfileStore` and fills `options` fields that were not
+/// already set on the CLI. Supports both the legacy single-profile YAML shape
```
Yaraslaut, line 448: I do not think that we have any legacy format here since we are not using it at the moment

Yaraslaut (changes requested):
I think that all additions to the `src/Lightweight` should actually go into `src/tools` and be compiled only as a part of the dbtool, and nothing for the library to provide

Yaraslaut (approved):
Thanks for the separation into smaller libraries

src/Lightweight/Config/ProfileStore.hpp
```diff
@@ -0,0 +1,168 @@
...
+    {
+        return _profiles;
+    }
+
+    /// Looks up a profile by name. Returns nullptr if none matches.
+    [[nodiscard]] Profile const* Find(std::string_view name) const noexcept;
+
+    /// Returns the default profile (by `DefaultProfileName()`), or the
+    /// first profile if no default is set, or nullptr if empty.
+    [[nodiscard]] Profile const* Default() const noexcept;
```
Yaraslaut, line 139: while I have a mixed feeling about iterators from functions like `find` i find it useful that they are distinct from a raw pointer and give you a way to check whether it is zero or not, maybe we can do something here similar, return a wrapper to the `profile*` and you need to use some monads to unwrap it

src/Lightweight/Secrets/backends/FileBackend.hpp
```diff
@@ -0,0 +1,69 @@
...
+    [[nodiscard]] std::optional<std::string> Read(std::string_view key) override;
+
+    /// Inserts or updates the secret for `key`. Rewrites the file atomically
+    /// (write-then-rename) with mode 0600 on POSIX. Throws if the parent
+    /// directory is not writable.
+    void Write(std::string_view key, std::string_view value) override;
+
+    /// Removes the entry for `key`, if present. Rewrites the file atomically.
+    void Erase(std::string_view key) override;
+
```
Yaraslaut, line 49: please provide strong types here with only explicit constructor from the `std::string_view`

src/Lightweight/Secrets/ISecretBackend.hpp
```diff
@@ -0,0 +1,64 @@
...
+    [[nodiscard]] virtual std::optional<std::string> Read(std::string_view key) = 0;
+
+    /// Stores `value` under `key`. Not every backend supports writing
+    /// (`env:` is read-only at runtime, for example); those should throw
+    /// `std::runtime_error` with a descriptive message.
+    virtual void Write(std::string_view key, std::string_view value) = 0;
+
+    /// Removes the secret at `key`. Silently succeeds if the key is absent;
+    /// read-only backends should throw just like in `Write`.
+    virtual void Erase(std::string_view key) = 0;
```
Yaraslaut, line 59: please strong typing for `Key` and `Value`

src/Lightweight/CMakeLists.txt
```diff
@@ -115,6 +132,14 @@ set(SOURCE_FILES
     SqlBackup/Restore.cpp
     SqlBackup/SqlBackup.cpp
     SqlBackup/TableFilter.cpp
+
+    Config/ProfileStore.cpp
+
+    Secrets/SecretResolver.cpp
+    Secrets/backends/EnvBackend.cpp
+    Secrets/backends/FileBackend.cpp
+    Secrets/backends/StdinBackend.cpp
```
Yaraslaut: Why is it a part of the `Lightweight` library ?

src/Lightweight/SqlMigration.cpp
```diff
@@ -416,8 +459,88 @@ namespace
...
+            auto const comma = list.find(',', begin);
+            auto const end = comma == std::string_view::npos ? list.size() : comma;
+            result.emplace_back(list.substr(begin, end - begin));
+            if (comma == std::string_view::npos)
+                break;
+            begin = comma + 1;
+        }
+        return result;
+    }
+
```
Yaraslaut, line 617: https://en.cppreference.com/cpp/ranges/chunk_by_view

src/Lightweight/SqlSchemaDiff.cpp
```diff
@@ -0,0 +1,428 @@
...
+    struct LogicalType
+    {
+        LogicalKind kind {};
+        /// Declared character/byte size for sized types. 0 means "unbounded" (e.g. TEXT,
+        /// or a driver that reports a sentinel huge size for unsized text columns).
+        std::size_t size = 0;
+        /// Decimal precision (only meaningful for @ref LogicalKind::Decimal).
+        std::size_t precision = 0;
+        /// Decimal scale (only meaningful for @ref LogicalKind::Decimal).
+        std::size_t scale = 0;
```
Yaraslaut, line 55: can we find a way to have this members only when they are needed ?

src/Lightweight/SqlSchemaDiff.cpp
```diff
@@ -0,0 +1,428 @@
...
+                    return { .kind = LogicalKind::Smallint };
+                else if constexpr (std::is_same_v<T, VarBinary>)
+                    // PostgreSQL maps `VarBinary(N)` to `BYTEA` and loses the size on
+                    // round-trip; sqlite preserves it. Treat VarBinary as unsized for
+                    // logical equivalence.
+                    return { .kind = LogicalKind::VarBinary };
+                else
+                    static_assert(sizeof(T) == 0, "Unhandled SqlColumnTypeDefinition alternative");
+            },
+            type);
```
Yaraslaut, line 128: i guess `overloaded` functor will be a nicer looking implementation for this huge visitor

src/Lightweight/SqlSchemaDiff.cpp
```diff
@@ -0,0 +1,428 @@
...
+        auto join = [](std::vector<std::string> const& v) {
+            auto s = std::string {};
+            bool first = true;
+            for (auto const& name: v)
+            {
+                if (!first)
+                    s += ", ";
+                s += name;
+                first = false;
+            }
```
Yaraslaut, line 246: https://en.cppreference.com/cpp/ranges/join_with_view

src/Lightweight/SqlSchemaDiff.cpp
```diff
@@ -0,0 +1,428 @@
...
+        auto join = [](std::vector<std::string> const& v) {
+            auto s = std::string {};
+            bool first = true;
+            for (auto const& name: v)
+            {
+                if (!first)
+                    s += ",";
+                s += name;
+                first = false;
+            }
```
Yaraslaut, line 268: https://en.cppreference.com/cpp/ranges/join_with_view

src/Lightweight/SqlDataDiff.cpp
```diff
@@ -0,0 +1,314 @@
...
+                return { '[', ']' };
+            case SqlServerType::MYSQL:
+                return { '`', '`' };
+            case SqlServerType::POSTGRESQL:
+            case SqlServerType::SQLITE:
+            case SqlServerType::UNKNOWN:
+                break;
+        }
+        return { '"', '"' };
+    }
```
Yaraslaut, line 35: this should be somewhere in the core parts of the library

src/Lightweight/SqlDataDiff.cpp
```diff
@@ -0,0 +1,314 @@
...
+
+} // namespace
+
+TableDataDiff DiffTableData(SqlConnection& a,
+                            SqlConnection& b,
+                            Table const& tableA,
+                            Table const& tableB,
+                            std::size_t maxRows,
+                            DiffProgressCallback onProgress)
+{
```
Yaraslaut, line 234: I think this should be done in parallel, we can compare initially the amount of rows in every database and then fetch and compare them in batches, it is important to fire queries that get few hundreds elements at once (depends on the max length of the sql query). So instead of `SqlConnection` it is better to provide a `DataMapperPool` for this database, such that this is save for parallel execution

Yaraslaut (changes requested):
Few comments, the most important is parallel diff for the databases

Yaraslaut:
this one as well, i think we should go back to this if needed

PR #486, opened 2026-05-05
Fix #485: SqlFixedString round-trip truncates values that fill capacity

### LINKED ISSUE #485: SqlFixedString: BufferLength=N passed to SQLBindCol/SQLGetData truncates values that fill the buffer
## Summary

`SqlBasicStringOperations<SqlFixedString<N>>::OutputColumn` and `::GetColumn` pass `BufferLength = N` to `SQLBindCol` / `SQLGetData`. Per the ODBC spec, `BufferLength` for `SQL_C_CHAR` must include space for the null terminator, so the driver truncates any value of length N to N − 1 data chars + null. Values shorter than N (with trailing whitespace padding from a `CHAR(N)` column) are unaffected because there is slack space.

## Code

`src/Lightweight/DataBinder/BasicStringBinder.hpp`, `OutputColumn`:

```cpp
if constexpr (requires { AnsiStringType::Capacity; })
    StringTraits::Resize(result, AnsiStringType::Capacity);   // _size = N
...
return SQLBindCol(stmt, column, SQL_C_CHAR,
                  (SQLPOINTER) StringTraits::Data(result),
                  (SQLLEN) StringTraits::Size(result),         // <-- BufferLength = N
                  indicator);
```

`SqlFixedString<N>` actually allocates its buffer as `T _data[N + 1]` (so the null-terminator slot exists), but the binder hands ODBC only `N` bytes.

The same pattern exists in the `GetColumn` overload (~line 291): `SQLGetData(..., AnsiStringType::Capacity, ...)` passes `N` instead of `N + 1`.

## Reproducer

Pseudocode in Lightweight test style:

```cpp
TEST_CASE_METHOD(SqlTestFixture,
                 "SqlTrimmedFixedString<3> round-trips a CHAR(3) value that fills the buffer",
                 "[SqlFixedString][regression]")
{
    auto stmt = SqlStatement{};
    stmt.ExecuteDirect("CREATE TABLE Currencies (NR INT, CODE CHAR(3))");
    stmt.ExecuteDirect("INSERT INTO Currencies VALUES (1, 'EUR')");

    SqlTrimmedFixedString<3> code;
    SQLLEN indicator = 0;
    // Bind code as output column 1, fetch the row, run post-processing.
    // Expected: code == "EUR", code.size() == 3
    // Actual:   code == "EU",  code.size() == 2
    REQUIRE(code.str() == "EUR");
    REQUIRE(code.size() == 3);
}
```

The existing `SqlFixedString: TrimRight` test in `tests/DataBinderTests.cpp` does not exercise this case — both calls pass `indicator < N` (e.g. `SqlTrimmedFixedString<20>{ "Hello        " }`, `TrimRight(&str, 5)`), so they have plenty of slack and don't reach the `min(indicator, N-1)` boundary.

## Downstream consequence in `TrimRight`

`SqlFixedString.hpp` (around line 474):

```cpp
size_t n = (std::min) (static_cast<size_t>(indicator) / sizeof(CharType), N - 1);
```

The `N - 1` cap here is consistent with the truncated buffer (since `SQLBindCol` could only have written `N − 1` data chars). It is therefore not separately observable today. Once `BufferLength` is corrected to `N + 1`, this cap also needs to become `N` — otherwise `_size` would still be capped at `N − 1` even though the buffer now has the full value.

## Suggested fix

In `OutputColumn` and `GetColumn` (and `BatchOutputColumn` / `Reserve` / `Resize` paths if symmetric), pass `Capacity + 1` instead of `Size(result)` / `Capacity` for types with a compile-time `Capacity`. Change the `N - 1` cap in `TrimRight` to `N`. Add a regression test where the stored value has `indicator == N` and the last byte is non-whitespace.

## Why this hasn't been observed before

The bug fires only when:

1. The stored value is exactly N chars, **and**
2. The last char is not whitespace (or null)

For most consumers of `SqlTrimmedFixedString<N>` with N ≥ 10, real values rarely fill the buffer with no padding. We tripped over it in a downstream codebase (LASTRADA, see internal ticket DEV-6285) via `SqlTrimmedFixedString<3>` mapped to a `WAEHRUNGEN.NAME CHAR(3)` column storing 3-letter ISO 4217 codes (EUR, CHF, GBP, USD, …). These come back as 2-char prefixes ("EU", "CH", "GB", "US"), silently breaking ISO 4217 lookup.

StepKie:
Closes #485.

## Problem

`BasicStringBinder<...>::OutputColumn` and `::GetColumn` were passing `BufferLength = Capacity` (i.e., `N`) to `SQLBindCol` / `SQLGetData` when binding an `SqlFixedString<N>`-derived field. ODBC's `SQL_C_CHAR` semantics require `BufferLength` to include space for the null terminator, so the driver truncated any value of length `N` to `N − 1` data chars + null.

`SqlFixedString<N>` already backs its data with `_data[N + 1]` so the terminator slot exists; the binder just wasn't telling ODBC about it.

The bug was masked for almost every consumer because real-world values rarely fill the full capacity with no trailing whitespace. We tripped over it in a downstream codebase (LASTRADA, ticket DEV-6285) via `Light::Field<Light::SqlTrimmedFixedString<3>>` mapped to a `WAEHRUNGEN.NAME CHAR(3)` column storing 3-letter ISO 4217 codes (`EUR`, `CHF`, `GBP`, `USD`, …). The codes came back as 2-char prefixes (`"EU"`, `"CH"`, `"GB"`, `"US"`), silently breaking ISO 4217 lookup and producing an empty `cbc:DocumentCurrencyCode` in XRechnung exports for non-Euro invoices.

## Approach

TDD — failing test in commit 1, fix in commit 2.

### Test (commit 1, `e5ebfc01`)

Adds `Trimmed fixed string preserves a value that fills its capacity` next to the existing `UnicodeTrimmedFixedRow` test in `DataBinderTests.cpp`. It uses `SqlTrimmedFixedString<3>` and the value `"EUR"` to hit `indicator == N` with a non-whitespace last byte — the precise condition the existing `SqlFixedString: TrimRight` test does not cover (its calls pass `indicator < N` and so always have slack).

### Fix (commit 2, `3a29c5b2`)

Two binder paths and one trim path:

1. `BasicStringBinder.hpp::OutputColumn` — for fixed-capacity types, pass `Capacity + 1` to `SQLBindCol` instead of `Size(result)`. Dynamic strings keep their existing behaviour.
2. `BasicStringBinder.hpp::GetColumn` (the `if constexpr (requires { Capacity; })` branch) — pass `Capacity + 1` to `SQLGetData`.
3. `SqlFixedString.hpp::TrimRight` — change the cap on `n` from `N − 1` to `N`. The previous `N − 1` was consistent with the truncated buffer (since ODBC could only have written `N − 1` data chars anyway). With `BufferLength` corrected, capping at `N − 1` would leave `_size` permanently one short of the actual content.

Existing tests are unaffected because they all operate on values shorter than `N` with trailing whitespace, where both `N − 1` and `N` give the same trim result.

## Test matrix

I haven't been able to run Lightweight's full test suite locally (no preconfigured ODBC test DB on this box). CI on this PR will exercise both the new regression test and the rest of the suite across the supported drivers. Happy to iterate if anything else trips.

### DIFF
--- a/src/Lightweight/DataBinder/BasicStringBinder.hpp
+++ b/src/Lightweight/DataBinder/BasicStringBinder.hpp
@@ -286,11 +286,22 @@ struct SqlDataBinder<AnsiStringType>
             cb.PlanPostProcessOutputColumn(
                 [stmt, column, indicator, result]() { PostProcessOutputColumn(stmt, column, result, indicator); });
 
+        // Per ODBC spec, BufferLength for SQL_C_CHAR must include space for the
+        // null terminator. SqlFixedString<N> backs its data with _data[N+1] for
+        // exactly this reason; pass Capacity+1 to let the driver write a full
+        // N-char value plus null. Dynamic strings keep the current behaviour.
+        SQLLEN const bufferLength = [result]() -> SQLLEN {
+            if constexpr (requires { AnsiStringType::Capacity; })
+                return static_cast<SQLLEN>(AnsiStringType::Capacity) + 1;
+            else
+                return static_cast<SQLLEN>(StringTraits::Size(result));
+        }();
+
         return SQLBindCol(stmt,
                           column,
                           SQL_C_CHAR,
                           (SQLPOINTER) StringTraits::Data(result),
-                          (SQLLEN) StringTraits::Size(result),
+                          bufferLength,
                           indicator);
     }
 
@@ -338,8 +349,10 @@ struct SqlDataBinder<AnsiStringType>
         if constexpr (requires { AnsiStringType::Capacity; })
         {
             StringTraits::Resize(result, AnsiStringType::Capacity);
+            // BufferLength for SQL_C_CHAR must include space for the null terminator;
+            // _data is sized N+1 to accommodate it.
             SQLRETURN const rv =
-                SQLGetData(stmt, column, SQL_C_CHAR, StringTraits::Data(result), AnsiStringType::Capacity, indicator);
+                SQLGetData(stmt, column, SQL_C_CHAR, StringTraits::Data(result), AnsiStringType::Capacity + 1, indicator);
             if (rv == SQL_SUCCESS || rv == SQL_NO_DATA)
             {
                 if (*indicator == SQL_NULL_DATA)
--- a/src/Lightweight/DataBinder/BasicStringBinder.hpp
+++ b/src/Lightweight/DataBinder/BasicStringBinder.hpp
@@ -290,19 +290,14 @@ struct SqlDataBinder<AnsiStringType>
         // null terminator. SqlFixedString<N> backs its data with _data[N+1] for
         // exactly this reason; pass Capacity+1 to let the driver write a full
         // N-char value plus null. Dynamic strings keep the current behaviour.
-        SQLLEN const bufferLength = [result]() -> SQLLEN {
+        SQLLEN const bufferLength = [&]() -> SQLLEN {
             if constexpr (requires { AnsiStringType::Capacity; })
                 return static_cast<SQLLEN>(AnsiStringType::Capacity) + 1;
             else
                 return static_cast<SQLLEN>(StringTraits::Size(result));
         }();
 
-        return SQLBindCol(stmt,
-                          column,
-                          SQL_C_CHAR,
-                          (SQLPOINTER) StringTraits::Data(result),
-                          bufferLength,
-                          indicator);
+        return SQLBindCol(stmt, column, SQL_C_CHAR, (SQLPOINTER) StringTraits::Data(result), bufferLength, indicator);
     }
 
     static void PostProcessOutputColumn(SQLHSTMT stmt, SQLUSMALLINT column, AnsiStringType* result, SQLLEN* indicator)
--- a/src/tests/DataBinderTests.cpp
+++ b/src/tests/DataBinderTests.cpp
@@ -1453,9 +1453,8 @@ TEST_CASE_METHOD(SqlTestFixture, "Trimmed fixed strings strip trailing padding t
 // Regression for #485: BasicStringBinder::OutputColumn passes BufferLength=N
 // (not N+1) to SQLBindCol, so ODBC truncates any value of length N to N-1
 // data chars + null. The bug only surfaces when the stored value fills the
-// full capacity with no trailing whitespace -- typical case is 3-letter
-// ISO 4217 codes in a CHAR(3) column. Surfaced downstream as LASTRADA DEV-6285
-// (empty cbc:DocumentCurrencyCode in XRechnung exports for non-Euro invoices).
+// full capacity with no trailing whitespace -- e.g. a 3-letter currency code
+// stored in a CHAR(3) column.
 struct FullCapacityFixedRow
 {
     Field<uint64_t, PrimaryKey::ServerSideAutoIncrement> id {};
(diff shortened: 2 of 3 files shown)


src/Lightweight/DataBinder/BasicStringBinder.hpp
```diff
@@ -286,11 +286,22 @@ struct SqlDataBinder<AnsiStringType>
...
+        }();
+
         return SQLBindCol(stmt,
                           column,
                           SQL_C_CHAR,
                           (SQLPOINTER) StringTraits::Data(result),
-                          (SQLLEN) StringTraits::Size(result),
+                          bufferLength,
                           indicator);
     }
```
StepKie, line 301: This _if -else lambda construction_ does not strike me as very elegant.

this is just what was generated by Claude. Feel free to replace if you find a solution less offensive to the eye that still passes the test.

src/tests/DataBinderTests.cpp
```diff
@@ -1450,6 +1450,35 @@ TEST_CASE_METHOD(SqlTestFixture, "Trimmed fixed strings strip trailing padding t
     CHECK(result->stringWide == SqlTrimmedWideFixedString<32> { L"Hellö" });
 }
 
+// Regression for #485: BasicStringBinder::OutputColumn passes BufferLength=N
+// (not N+1) to SQLBindCol, so ODBC truncates any value of length N to N-1
+// data chars + null. The bug only surfaces when the stored value fills the
+// full capacity with no trailing whitespace -- typical case is 3-letter
+// ISO 4217 codes in a CHAR(3) column. Surfaced downstream as LASTRADA DEV-6285
+// (empty cbc:DocumentCurrencyCode in XRechnung exports for non-Euro invoices).
```
Yaraslaut: internal ticket mentioned in public repo

Yaraslaut (commented):
Also, this is quite surprising that we get this issue,  is it something specific to the trimmed strings? since we have plenty of tests for other string types

src/Lightweight/DataBinder/BasicStringBinder.hpp
Yaraslaut: this is the pattern we use in the code to make variables `const`

Yaraslaut (approved):

StepKie:
> Also, this is quite surprising that we get this issue, is it something specific to the trimmed strings? since we have plenty of tests for other string types

@Yaraslaut Not specific to trimmed mode — `BufferLength = N` is passed in `OutputColumn` regardless of `SqlFixedStringMode`. The reason it slipped through is that no test in the suite stores a value that fills the full capacity with a non-whitespace last byte:

- `SqlFixedString: TrimRight` calls `TrimRight(&str, 5)` on `SqlTrimmedFixedString<20>` / `<32>` — `indicator = 5` vs `N = 20`, so 15 bytes of slack and the cap at `N − 1` never bites.
- `Trimmed fixed strings strip trailing padding through DataMapper` stores `"Hello"` (5 chars) in a `<32>` field — same: plenty of slack.

Nothing in the suite probes `indicator == N` with a non-whitespace last byte. Trimmed mode is just where it's easiest to *spot* (round-trip `"EUR"` → get `"EU"`) — for variable-size and non-trimmed fixed mode the same off-by-one chops the last byte too, it just looks less suspicious without an exact-length probe.

🤖 ai-assisted

StepKie:
Apologies @Yaraslaut — when I amended just now to drop the internal ticket reference from the test comment, I force-pushed from a local branch that pre-dated your `[clang-tidy] small fixes` commit (`a05eb9ab`), so I silently overwrote your `[result]` → `[=]` change. `--force-with-lease` didn't catch it because my tracking branch was stale (I hadn't fetched).

The latest commit (`08147c2e`) re-applies an equivalent fix: `[result]` → `[&]` (capture-all-by-reference rather than by-value, otherwise same effect — silences `-Wunused-lambda-capture` under `-Werror` regardless of which `if constexpr` branch is selected, and matches the IIFE style used elsewhere in the codebase).

Sorry for the noise. CI should go back green.

🤖 ai-assisted

PR #489, opened 2026-05-13
Fix MSSQL "Connection is busy" during long migration runs

christianparpart:
Applying the full Lastrada migration set against MS SQL Server (411 migrations through `dbtool-gui`, or 356 through `dbtool`) intermittently failed with `HY000 [Microsoft][ODBC Driver 18 for SQL Server]Connection is busy with results for another command` on routine DDL like `IF NOT EXISTS (SELECT … FROM sys.columns …) ALTER TABLE … ADD …;`. Two independent root causes were involved — the CLI hit one of them, the GUI hit both.

## Changes

- **`SqlStatement::CloseCursor` now drains pending result sets via `SQLMoreResults` before `SQLFreeStmt(SQL_CLOSE)`.** SQL Server emits per-statement row-count / DONE_IN_PROC tokens for every statement in a T-SQL batch; with MARS off these stay pending on the connection and poison the next `SQLExecDirect`. `SQLMoreResults` is standard ODBC — SQLite and PostgreSQL drivers return `SQL_NO_DATA` on the first call, so the cost on single-statement queries is one no-op driver call.
- **`SET NOCOUNT ON;` on the SQL Server `sp_getapplock` / `sp_releaseapplock` batches** (defense in depth so neither path can leave tokens pending even if a future caller forgets to drain).
- **`AppController::currentReleaseLabel` no longer hits the database from the UI thread.** The original implementation called `MigrationManager::GetAppliedMigrationIds()` synchronously on QML binding evaluation, which fires on every `migrationsChanged()` signal — including the one emitted between every migration during an apply run. That `SELECT` raced the worker thread's `ApplySingleMigration` on the shared `SqlConnection`, surfacing as HY000 regardless of MARS or pending result sets. It now derives the applied set from the in-memory `MigrationListModel`, which is already kept in sync per row by `SetRowStatusByTimestamp` during runs.
- **`previewMigrationSql` short-circuits during a run.** `PreviewMigration` issues `INFORMATION_SCHEMA` queries via `MakeWidthLookup` and would hit the same race; this gates it on `_runner.phase() == Idle`.
- New `MigrationListModel::AppliedTimestamps()` accessor + two MSSQL-gated regression tests for the SQLMoreResults path.

The underlying `SqlConnection` remains not thread-safe; a structural fix (separate read-only UI connection or a connection-level mutex) is left as future work.

### DIFF
--- a/src/Lightweight/SqlQueryFormatter.cpp
+++ b/src/Lightweight/SqlQueryFormatter.cpp
@@ -159,7 +159,13 @@ namespace
                                                                    std::string_view lockName,
                                                                    std::chrono::milliseconds timeout) const override
         {
-            auto const sql = std::format("DECLARE @result INT; "
+            // SET NOCOUNT ON suppresses the per-statement row-count "DONE"
+            // tokens that would otherwise surface as additional result sets
+            // on top of the SELECT @result we actually consume. Combined with
+            // SqlStatement::CloseCursor draining via SQLMoreResults, this
+            // guarantees no pending token survives the lock acquire.
+            auto const sql = std::format("SET NOCOUNT ON; "
+                                         "DECLARE @result INT; "
                                          "EXEC @result = sp_getapplock @Resource = N'{}', @LockMode = N'Exclusive', "
                                          "@LockTimeout = {}, @LockOwner = N'Session'; "
                                          "SELECT @result;",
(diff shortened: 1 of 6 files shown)


StepKie:
That is some serious AI gobledigook in the description ;)

christianparpart:
@StepKie haha yeah, I however like it, because I want @Yaraslaut to understand why I did that. I now ran twice into the issue of running all of our company internal SQL migrations on a MS SQL instance, and then getting these errors. Next week I want to introduce the GUI(/CLI) db & migration tool to the team.

PR #491, opened 2026-05-14
Add `dbtool list-profiles` subcommand

christianparpart:
Adds a `list-profiles` subcommand to `dbtool` so operators can discover which connection profiles their config file (`~/.config/dbtool/dbtool.yml` or `%APPDATA%\dbtool\dbtool.yml`) defines without opening the YAML by hand. The command does not open a database connection, so it dispatches early alongside `resolve-secret` / `help` / `show-examples` and works on any platform without an ODBC driver, a plugin directory, or a reachable database.

Output is a compact table (NAME / DEFAULT / CONNECTION / SCHEMA / PLUGINSDIR). `PWD=` / `Password=` values inside a profile's raw `connectionString` are redacted to `***` before display — the YAML never persists secrets via `secretRef`, but users sometimes paste passwords directly into `connectionString` and the listing should not echo them back to a shared terminal.

A missing `--config <FILE>` is a hard error (matches the existing `ApplyProfileToOptions` behaviour); a missing default-path config emits a friendly notice and exits 0 so a fresh install with no config file explains itself instead of looking broken.

## Changes

- `src/tools/dbtool/main.cpp`: new `RedactConnectionStringSecrets()` helper (case-insensitive, `;`-terminated, anchored at start-of-string or after `;`); new `ListProfiles()` handler; early dispatch in `main()`; `Commands:` and `Examples:` entries in the help text.
- `src/tests/test_dbtool.py`: integration step `12a` that writes a temp YAML with two profiles (one carrying `PWD=secret`), runs the new command, and asserts both names appear, exactly one row carries the default marker, the password is absent from the output, `PWD=***` is present, and `--config <missing>` exits non-zero.
- `docs/dbtool.md`: "Inspecting configured profiles" subsection under Configuration with example output.

### DIFF
--- a/src/tools/dbtool/main.cpp
+++ b/src/tools/dbtool/main.cpp
@@ -12,6 +12,9 @@
 #include <Lightweight/SqlSchema.hpp>
 #include <Lightweight/SqlScopedLock.hpp>
 
+#include <algorithm>
+#include <array>
+#include <cctype>
 #include <cstdio>
 #include <expected>
 #include <filesystem>
@@ -162,6 +165,8 @@ void PrintUsage()
     std::println("  {}restore{} --input FILE     Restores the database from a file", c.command, c.reset);
     std::println("  {}resolve-secret{} {}<REF>{}   Prints a resolved secret to stdout (env:, file:, stdin:)",
                  c.command, c.reset, c.param, c.reset);
+    std::println("  {}list-profiles{}            Lists profiles from the configuration file (see --config)",
+                 c.command, c.reset);
     std::println("");
 
     // Descriptions start at column 29 (longest option is 27 chars + 2 space minimum gap)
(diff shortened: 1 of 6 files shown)


src/tools/dbtool/main.cpp
```diff
@@ -770,6 +824,133 @@ bool SetupConnectionString(SqlConnectionString const& connectionString)
...
+        std::println(std::cerr, "Error loading config file: {}", storeResult.error());
+        return EXIT_FAILURE;
+    }
+    auto const& store = *storeResult;
+
+    if (store.Empty())
+    {
+        std::println("No profiles configured in {}.", configPath.string());
+        return EXIT_SUCCESS;
+    }
```
Yaraslaut: I think that all of this should be inside lambda, that return std::expected with the std::expected<ProfileStore, std::message>`

src/tools/dbtool/main.cpp
```diff
@@ -770,6 +824,133 @@ bool SetupConnectionString(SqlConnectionString const& connectionString)
...
+        auto width = header.size();
+        for (auto const& row: rows)
+            width = std::max(width, selector(row).size());
+        return width;
+    };
+
+    auto const nameW = widthOf([](Row const& r) -> std::string const& { return r.name; }, "NAME");
+    auto const defW = widthOf([](Row const& r) -> std::string const& { return r.defaultMarker; }, "DEFAULT");
+    auto const connW = widthOf([](Row const& r) -> std::string const& { return r.connection; }, "CONNECTION");
+    auto const schemaW = widthOf([](Row const& r) -> std::string const& { return r.schema; }, "SCHEMA");
```
Yaraslaut: better to use projector here, it will be less boilerplate

Yaraslaut (commented):
Oh, great addition, thanks a lot :)

PR #492, opened 2026-05-17
[CI] Add various sql configuration for the tests

### LINKED ISSUE #478: Extend CI tests
Make sure that we run tests against all existing drivers on windows, and also when compiled on windows and not in linux environment

Yaraslaut:
Closes https://github.com/LASTRADA-Software/Lightweight/issues/478


Yaraslaut:
@christianparpart  Please see added CI tests, as you see we have quite a few unit tests failing for the legacy driver, what we should do about it?

PR #497, opened 2026-05-19
dbtool: defaultPluginsDir as list of dirs; DI-friendly plugin ingestion

christianparpart:
`dbtool.yml`'s `defaultPluginsDir` now accepts a YAML sequence in addition to the existing scalar form, so plugins can be spread across several directories (a vendor-shipped baseline + a local override directory, say). When the same plugin filename appears in more than one of the listed directories, the file with the newest modification time wins; ties go to the directory listed earlier. Filename comparison is case-insensitive on Windows and case-sensitive elsewhere. A new `--verbose` (`-v`) flag on both `dbtool` and `dbtool-gui` prints one informational line per shadowed plugin so the dedup decision is debuggable.

The discover + load + merge orchestration is extracted into a single shared helper `Lightweight::Tools::IngestPlugins` (in `tools_shared/`), parameterised on a `PluginLoaderFn` callback so unit tests can substitute a fake for the real `PluginLoader`. This deletes the duplicated `LoadPlugins` + `CollectMigrations` pair from `dbtool/main.cpp` and the `LoadPluginsInto` helper from `dbtool-gui/AppController.cpp`; both binaries now go through the exact same orchestration and cannot drift.

## Changes

- `defaultPluginsDir` accepts either a string or a YAML sequence; both forms round-trip through save.
- New `--verbose` / `-v` CLI flag on `dbtool` and `dbtool-gui` that surfaces shadowed-plugin notices via stderr / `qInfo()`.
- New shared `Lightweight::Tools::DiscoverPlugins` helper performs the filename-based dedup (newest mtime wins; case-insensitive on Windows).
- New shared `Lightweight::Tools::IngestPlugins(dirs, loader, central, options)` plus `DefaultPluginLoader()` — the orchestration is now DI-friendly via the loader callback and the per-channel logging callbacks (`logShadowed`, `logLoading`, `logError`).
- `PluginLoader` moved from `src/tools/dbtool/` to `src/tools/shared/` since both binaries already shared it; `dbtool-gui` no longer reaches into `../dbtool/` for the include.
- `dbtool-gui`'s plugin storage is pimpl'd (`PluginsBundle` defined in the `.cpp`) so the moc-parsed `AppController.hpp` does not pull in `<PluginIngestion.hpp>`.
- 10 new unit tests for `IngestPlugins` covering loader-call-count, nullopt-skip, throw-then-rethrow, throw-then-continue, duplicate-handle reject, releases + migrations + compat-policy propagation, `pluginManager == &central` no-op, `logShadowed` / `logLoading` wiring, and `LoadedPlugin::TryGetFunction` symbol lookup.
- 6 new unit tests for `DiscoverPlugins` covering distinct names, mtime-wins, mtime-tie + earlier-dir, non-plugin extension skip, missing-dir skip, and platform-specific case sensitivity.
- 4 new `ConfigProfileStoreTests` for scalar/sequence YAML parse, malformed sequence rejection, and multi-entry round-trip.
- New `test_dbtool.py` step verifying two dirs with same-named plugins + `--verbose` end-to-end against a real `dummy_migration_plugin.dll`.
- Docs update in `docs/dbtool.md` covering the list form, the dedup rule (newest mtime, ties go to earlier dir), and the `--verbose` output.

Also includes a separate WIP commit adding CPack packaging (WIX installer on Windows, TGZ/DEB/RPM on Linux) that was already in flight on the working tree.


docs/dbtool.md
```diff
@@ -23,6 +40,22 @@ cmake --build out/build/clang-release --target dbtool
 
 The binary will be located at `out/build/clang-release/src/tools/dbtool/dbtool`.
 
+### Producing the installer locally
+
+```bash
+# Linux (TGZ + DEB + RPM)
+cmake --preset gcc-release
+cmake --build --preset gcc-release
+cpack --preset gcc-release
```
Yaraslaut, line 49: I think that we also have something like 
`cmake --preset gcc-release --install` and it allows installing only one component at a time, for instance:
```
cmake --preset gcc-release --install --component dbtool-gui
```

docs/dbtool.md
```diff
@@ -58,9 +91,33 @@ profiles:
...
+  - /opt/lightweight/plugins        # vendor-shipped baseline
+```
+
+When the same plugin filename appears in more than one of those directories,
+dbtool keeps the file with the newest modification time and discards the
+others — so dropping a fresher build into `./plugins` reliably shadows the
+baseline copy without having to rebuild the YAML. Filename comparison is
+case-insensitive on Windows and case-sensitive elsewhere; when two
+same-named files share the same modification time, the directory listed
+earlier in the sequence wins.
```
Yaraslaut, line 110: this is a nice addition

src/tools/dbtool-gui/AppController.hpp
```diff
@@ -205,6 +202,16 @@ class AppController: public QObject
...
+
+    /// Seeds the initial value of `_verbose` before the QML engine
+    /// instantiates the singleton. Must be called before any QML access to
+    /// `AppController`. Mirrors `SeedBackupRestoreEnabled` so `main.cpp` can
+    /// thread the parsed `--verbose`/`-v` flag into the controller's first
+    /// `ReloadPlugins()`. Idempotent; later changes take effect on the next
+    /// `ReloadPlugins()` only if the value is updated via `SetVerbose`.
+    /// @param enabled Whether shadow-plugin notices should be emitted via
+    ///                `qInfo()`.
+    static void SeedVerbose(bool enabled) noexcept;
```
Yaraslaut, line 221: i am always having issues with such signatures,  what is `SeedVerbose(false)` means? is it equivalent to not calling this function? and if yes, then when should i call this function with `false` as an argument

src/tools/shared/PluginDiscovery.cpp
```diff
@@ -0,0 +1,205 @@
...
+    // Plain index loop because `std::views::enumerate` on a
+    // `std::span<T const>` does not compile on every libc++ version we
+    // target (the C++26-reflection CI job uses an older libc++).
+    for (std::size_t dirIndex = 0; dirIndex < dirs.size(); ++dirIndex)
+        ScanOneDirectory(dirs[dirIndex], dirIndex, encounter, winners, logShadowed);
+
+    std::vector<ResolvedPlugin> result;
+    result.reserve(winners.size());
+    for (auto const& [_, c]: winners)
+        result.push_back(ResolvedPlugin { .path = c.path, .dirIndex = c.dirIndex });
```
Yaraslaut: emplace_back

Yaraslaut (approved):
only small points, looked mainly only at headers

docs/dbtool.md
christianparpart: cmake --install is just doing `--target install`, which installs to `DESTDIR` and not actually creating installable packages.

PR #499, opened 2026-05-20
Extend query builder to not allow illformed query generation

### LINKED ISSUE #453: Sql query builder
* given the query, you should be able to construct sql string on const query
```
auto const query = dm.FromTable()....;
query.ToSql();
```

* Fields() function must accept also SqlQualifiedTableColumnName

Yaraslaut:
Closes https://github.com/LASTRADA-Software/Lightweight/issues/453 '

Fixes ill formed query builder API, where we could have `Select().All()` 

We split the select builder into two builders, where you can not complete the starter, before you select fields using `.Fields` overload

### DIFF
--- a/src/Lightweight/SqlQuery/Select.hpp
+++ b/src/Lightweight/SqlQuery/Select.hpp
@@ -9,6 +9,7 @@
 #include <reflection-cpp/reflection.hpp>
 
 #include <span>
+#include <utility>
 
 namespace Lightweight
 {
@@ -85,8 +86,16 @@ namespace Aggregate
 
 /// @brief Query builder for building SELECT ... queries.
 ///
+/// Not constructed directly by user code; obtained by adding a projection
+/// (`Field` / `Fields` / `FieldAs` / `Build`) to a @ref SqlSelectQueryStarter — the
+/// type returned by `SqlQueryBuilder::Select()`. The starter intentionally hides
+/// the finalizers @ref All, @ref First, and @ref Range so that an empty
+/// `Select().All()` (which would emit malformed `SELECT  FROM "T"` SQL) fails at
+/// compile time; the gate is in the starter, not here.
+///
 /// @see SqlQueryBuilder
-class [[nodiscard]] SqlSelectQueryBuilder final: public SqlBasicSelectQueryBuilder<SqlSelectQueryBuilder>
+/// @see SqlSelectQueryStarter
+class [[nodiscard]] SqlSelectQueryBuilder: public SqlBasicSelectQueryBuilder<SqlSelectQueryBuilder>
 {
   public:
     /// The select query type alias.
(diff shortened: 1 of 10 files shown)


src/tests/CoreTests.cpp
```diff
@@ -735,4 +735,45 @@ TEST_CASE("ParseConnectionString tolerates empty fragments", "[ConnectionString]
...
+// produce malformed `SELECT  FROM "Employees"` SQL at runtime.
+//
+//   auto queryIllFormed = dm.FromTable("Employees").Select();
+//   auto const queryIllFormedComplete = queryIllFormed.All();  // compile error
+//
+//   auto const chained = dm.FromTable("Employees").Select().All();  // compile error
+
+TEST_CASE_METHOD(SqlTestFixture, "Query building", "[SqlStatement]")
+{
+
```
christianparpart: how come we end up with that one line here?

src/tests/CoreTests.cpp
```diff
@@ -735,4 +735,45 @@ TEST_CASE("ParseConnectionString tolerates empty fragments", "[ConnectionString]
...
+    auto stmt = Light::SqlStatement {};
+    CreateEmployeesTable(stmt);
+    FillEmployeesTable(stmt);
+
+    auto dm = Light::DataMapper {};
+
+    SECTION("Select().All()")
+    {
+
```
christianparpart: same

src/Lightweight/SqlQuery/Select.hpp
```diff
@@ -250,4 +279,157 @@ inline LIGHTWEIGHT_FORCE_INLINE SqlSelectQueryBuilder& SqlSelectQueryBuilder::Fi
...
+class [[nodiscard]] SqlSelectQueryStarter final: private SqlSelectQueryBuilder
+{
+  public:
+    /// Constructs a SELECT query starter.
+    explicit SqlSelectQueryStarter(SqlQueryFormatter const& formatter, std::string table, std::string tableAlias) noexcept:
+        SqlSelectQueryBuilder { formatter, std::move(table), std::move(tableAlias) }
+    {
+    }
+
+    // Re-expose accessors and the safe inherited methods.
```
christianparpart: i'm not sure I like what I see. Is this really the best software architecture in place to achieve this? Marking a base class privately inherited and then re-expose privately inherited members feels like abuse.

christianparpart (approved):
Thanks. This is definitely a great addition. I am only concerned about the software architectural decision on private inherited classes and re-exposed members. And the unnecessary farm of empty lines as commented :)

src/Lightweight/SqlQuery/Select.hpp
Yaraslaut: We need somehow to disallow calling `.All()` before calling `.Field()` so we need to insert one more layer in the chain but do not want to re implement functions, we can try to shadow function and have a public inheritance, maybe like this: https://godbolt.org/z/s4jWE3YK6 what do you think?

christianparpart:
Ah nice, thanks, @Yaraslaut

PR #500, opened 2026-05-20
tools: filter plugin candidates by '*Plugin' filename suffix

christianparpart:
## Summary

- `Lightweight::Tools::DiscoverPlugins` now rejects shared libraries whose file stem does not end in the literal token `Plugin` (e.g. `FooPlugin.dll`, `SqlMigrationsPlugin.so`). The filter runs **before** `LoadLibrary` / `dlopen`, so neighbouring runtime DLLs (vcpkg-applocal deposits like `Lightweight.dll` / `zlibd1.dll` / `bz2d.dll`, Qt6*.dll dropped next to a GUI tool, MSVC runtimes, …) are never mapped into the host process and never run their `DllMain`.
- Case-insensitive on Windows (matching the filesystem and `LoadLibraryW`); case-sensitive on POSIX. Mirrors the existing `FilenameKeyFromPath` policy.
- The `TryGetFunction("AcquireMigrationManager")` null-skip in `DefaultPluginLoader` is kept as defence-in-depth: a `*Plugin.dll` without the entry point is still silently skipped rather than aborting dbtool.

## Motivation

A debug build of a downstream project (Lastrada) puts the migration plugin DLL in `<build>/plugins/`. After link, vcpkg's applocal deployment step (`VCPKG_APPLOCAL_DEPS=ON`) copies the plugin's transitive runtime DLL closure into the same directory. Result: dbtool would `LoadLibrary` `Lightweight.dll` / `zlibd1.dll` / `bz2d.dll` / etc. on every startup, run their DllMain, walk their import tables, only to discard them seconds later for missing the plugin entry point. With the filter, only `SqlMigrationsPlugin.dll` is loaded.

## Knock-on changes

- The two dummy migration plugin targets (`src/tests/dummy_migration_plugin{,_2}/CMakeLists.txt`) set `OUTPUT_NAME` to `DummyMigrationPlugin` / `DummyMigrationPlugin2` and `PREFIX ""` so the built artefacts conform to the new convention on both Windows and POSIX. Source folders and CMake target names are unchanged.
- `PluginDiscoveryTests.cpp` / `PluginIngestionTests.cpp` fixture filenames renamed to conforming names.
- New `TEST_CASE("DiscoverPlugins — DLLs whose stem does not end in 'Plugin' are skipped")` mirrors the applocal-style clutter and asserts only the plugin survives discovery.

## Documentation

The naming requirement is spelled out in the leading doc block of `DiscoverPlugins` in `PluginDiscovery.hpp`, with a "why" pointing at the applocal scenario above.

## Test plan

- [x] Unit tests: `LightweightTest` (Catch2) — existing tests pass after fixture rename, new `applocal-like` test asserts the filter rejects the four neighbours.
- [x] dbtool integration test (`test_dbtool.py`) — enumerates the test plugins dir by extension; the `OUTPUT_NAME` rename is transparent. No script change needed.
- [x] Manual: smoke-test `dbtool list-pending` against a `plugins/` directory that contains `SqlMigrationsPlugin.dll` + `Lightweight.dll` + `zlibd1.dll` and confirm only the plugin is loaded.

### DIFF
--- a/src/tests/PluginDiscoveryTests.cpp
+++ b/src/tests/PluginDiscoveryTests.cpp
@@ -90,8 +90,8 @@ TEST_CASE("DiscoverPlugins — distinct filenames across directories all survive
     auto const dirA = tree.Dir("a");
     auto const dirB = tree.Dir("b");
 
-    Touch(dirA / "alpha.dll", BaseTime());
-    Touch(dirB / "beta.dll", BaseTime());
+    Touch(dirA / "AlphaPlugin.dll", BaseTime());
+    Touch(dirB / "BetaPlugin.dll", BaseTime());
 
     std::vector<std::filesystem::path> dirs { dirA, dirB };
     auto const resolved = Lightweight::Tools::DiscoverPlugins(dirs);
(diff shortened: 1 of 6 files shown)


Yaraslaut (approved):
looks good

PR #501, opened 2026-05-21
Improve dbtool-gui log pane with startup, plugin, and connection events

christianparpart:
Today `LogPanel` in dbtool-gui is fed only by `MigrationRunner::logLine`, so it sits empty until the user clicks Migrate. A user opening dbtool-gui cannot tell which plugins were loaded, which database was just connected, or where the profile store lives — even though all of that is already known internally. This PR makes the pane a chronological narrative of "what just happened".

## Changes

- **Startup banner** in `AppController`'s constructor: Qt version, theme (effective + requested), view mode, profile-store path + profile count.
- **Plugin discovery** (per `ReloadPlugins`): resolved plugins directory, per-plugin `Loaded` / `Shadowed` / `Failed` lines, summary count of migrations & releases. Runs unconditionally — `--verbose` no longer gates discovery output, it only controls extra stderr noise.
- **Database connect** (per `connectToProfile`): `Connecting <summary>…`, `Connected: <ServerType> · <database> · <version>`, optional `Schema search path: …`, and `Applied migrations: A / T`. Failures log `Connection failed: …` in addition to the existing error banner. Disconnect emits `Disconnected.`
- **Typed log level**: new `enum class DbtoolGui::LogLevel { Info, Warning, Error }` with `Q_NAMESPACE` + `Q_ENUM_NS` replaces every magic `0`/`1`/`2` at the C++ signal boundary (`MigrationRunner::logLine`, `BackupRunner::logLine`, and the new `AppController::logLine`).
- **Unified feed**: `AppController` gains a single `logLine` signal; the runners forward into it so QML has one subscription source. Messages emitted before `LogPanel.qml` exists (plugin discovery happens in the constructor, well before the QML engine wires up its receiver) are buffered and replayed when QML calls `attachLogSink()` from `Component.onCompleted`.
- **Reused, never reinvented**: `connectionSummary()` continues to redact secrets, so the pane never leaks passwords; `SqlConnection::{ServerType,DatabaseName,ServerVersion}` is queried post-handshake so the reported flavour matches what the driver actually negotiated.
- **No more stderr duplication**: the three `fprintf(stderr, ...)` startup lines in `main.cpp` are removed — their content is now in the pane, where users can actually see it.

No public API surface changes; `LogLevel` lives in the `DbtoolGui` namespace and is only used by the GUI binary. `LightweightTest` remains green on SQLite (7396 assertions pass); `dbtool` itself is untouched.

### DIFF
--- a/src/tools/dbtool-gui/AppController.cpp
+++ b/src/tools/dbtool-gui/AppController.cpp
@@ -6,9 +6,11 @@
 #include <Lightweight/SqlConnection.hpp>
 #include <Lightweight/SqlError.hpp>
 #include <Lightweight/SqlMigration.hpp>
+#include <Lightweight/SqlServerType.hpp>
 
 #include <PluginIngestion.hpp>
 #include <PluginLoader.hpp>
+#include <QtCore/QtGlobal>
 
 namespace DbtoolGui
 {
@@ -39,6 +41,8 @@ struct AppController::PluginsBundle
 #include <QtCore/QTimer>
 #include <QtCore/QUrl>
 #include <QtGui/QDesktopServices>
+#include <QtGui/QGuiApplication>
+#include <QtGui/QStyleHints>
 
 namespace DbtoolGui
 {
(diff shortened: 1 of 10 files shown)


Yaraslaut (approved):

christianparpart:
Folded into #502 — commit ddec8076 has been cherry-picked into the maintenance/lazy-schema-migrations branch so the log-pane wiring lands together with the schema_migrations overlay redesign. PR 502 now sources its post-init hook log messages through AppController::LogInfo / LogError, which is what makes the plugin status banner show up in the log pane during profile switches.

PR #505, opened 2026-06-10
[dbtool] Backup/Restore perf improvements

Yaraslaut:
some perf numbers for the `MS SQL`
```
  Backup

  ┌──────┬────────┬────────┬─────────┐
  │ jobs │ master │ branch │ speedup │
  ├──────┼────────┼────────┼─────────┤
  │ 1    │ 2.11 s │ 1.74 s │ 1.22×   │
  ├──────┼────────┼────────┼─────────┤
  │ 2    │ 2.11 s │ 0.97 s │ 2.17×   │
  ├──────┼────────┼────────┼─────────┤
  │ 4    │ 2.11 s │ 0.60 s │ 3.50×   │
  ├──────┼────────┼────────┼─────────┤
  │ 8    │ 2.09 s │ 0.48 s │ 4.35×   │
  └──────┴────────┴────────┴─────────┘

  Restore

  ┌──────┬────────┬────────┬─────────┐
  │ jobs │ master │ branch │ speedup │
  ├──────┼────────┼────────┼─────────┤
  │ 1    │ 6.77 s │ 6.80 s │ 0.99×   │
  ├──────┼────────┼────────┼─────────┤
  │ 4    │ 6.69 s │ 3.77 s │ 1.77×   │
  ├──────┼────────┼────────┼─────────┤
  │ 8    │ 6.84 s │ 2.21 s │ 3.10×   │
  └──────┴────────┴────────┴─────────┘


```

### DIFF
--- a/src/Lightweight/DataBinder/BasicStringBinder.hpp
+++ b/src/Lightweight/DataBinder/BasicStringBinder.hpp
@@ -7,7 +7,6 @@
 
 #include <cassert>
 #include <cstddef>
-#include <cstdint>
 #include <cstring>
 #include <memory>
 #include <optional>
@@ -43,6 +42,11 @@ namespace detail
     {
         using CharType = ArrayType::value_type;
 
+        // Character C types NUL-terminate every SQLGetData transfer, consuming the last buffer
+        // slot; SQL_C_BINARY transfers raw bytes WITHOUT a terminator, so every buffer slot holds
+        // data. Getting this wrong for binary loses the byte at each continuation boundary.
+        constexpr size_t TrailingNulTerminators = CType == SQL_C_BINARY ? 0 : 1;
+
         *indicator = 0;
 
         // Resize the string to the size of the data
@@ -64,14 +68,18 @@ namespace detail
             return sqlResult;
         }
 
-        if (sqlResult == SQL_SUCCESS_WITH_INFO && *indicator > static_cast<SQLLEN>(result->size()))
+        if (sqlResult == SQL_SUCCESS_WITH_INFO && *indicator != SQL_NO_TOTAL
+            && (static_cast<size_t>(*indicator) / sizeof(CharType)) + TrailingNulTerminators > result->size())
         {
+            // Truncation with a known total. Note the TrailingNulTerminators term: a CHARACTER value whose
+            // length exactly equals the buffer still truncates (the driver spends the last slot
+            // on the NUL terminator), so the continuation must run for indicator == buffer size.
             // We have a truncation and the server knows how much data is left.
             auto const totalCharCount = static_cast<size_t>(*indicator) / sizeof(CharType);
-            auto const charsWritten = result->size() - 1;
-            result->resize(totalCharCount + 1);
+            auto const charsWritten = result->size() - TrailingNulTerminators;
+            result->resize(totalCharCount + TrailingNulTerminators);
             auto* bufferCont = result->data() + charsWritten;
-            auto const bufferCharsAvailable = (totalCharCount + 1) - charsWritten;
+            auto const bufferCharsAvailable = (totalCharCount + TrailingNulTerminators) - charsWritten;
             sqlResult = SQLGetData(
                 stmt, column, CType, bufferCont, static_cast<SQLLEN>(bufferCharsAvailable * sizeof(CharType)), indicator);
             if (SQL_SUCCEEDED(sqlResult))
@@ -83,13 +91,17 @@ namespace detail
         while (sqlResult == SQL_SUCCESS_WITH_INFO && *indicator == SQL_NO_TOTAL)
         {
             // We have a truncation and the server does not know how much data is left.
-            writeIndex += result->size() - 1;
+            writeIndex += result->size() - TrailingNulTerminators;
             result->resize(result->size() * 2);
             auto* const bufferStart = result->data() + writeIndex;
             size_t const bufferCharsAvailable = result->size() - writeIndex;
             sqlResult = SQLGetData(
                 stmt, column, CType, bufferStart, static_cast<SQLLEN>(bufferCharsAvailable * sizeof(CharType)), indicator);
         }
+        // The unknown-total loop exits with the final transfer's char count in the indicator; trim
+        // the over-allocated buffer to the bytes actually written.
+        if (writeIndex > 0 && SQL_SUCCEEDED(sqlResult) && *indicator != SQL_NULL_DATA && *indicator != SQL_NO_TOTAL)
+            result->resize(writeIndex + (static_cast<size_t>(*indicator) / sizeof(CharType)));
         return sqlResult;
     }
 
--- a/src/Lightweight/DataBinder/BasicStringBinder.hpp
+++ b/src/Lightweight/DataBinder/BasicStringBinder.hpp
@@ -388,6 +388,62 @@ struct SqlDataBinder<AnsiStringType>
     static LIGHTWEIGHT_FORCE_INLINE SQLRETURN OutputColumn(
         SQLHSTMT stmt, SQLUSMALLINT column, AnsiStringType* result, SQLLEN* indicator, SqlDataBinderCallback& cb) noexcept
     {
+        // PostgreSQL: read narrow strings back through UTF-16 + SQL_C_WCHAR, mirroring InputParameter.
+        // psqlODBC on Windows transcodes SQL_C_CHAR data to the system codepage (cp1252), which mangles
+        // the UTF-8 bytes the write path stored (café -> caf? ?). Going via SQL_C_WCHAR and converting
+        // UTF-16 -> UTF-8 here keeps the narrow round-trip byte-exact under the library's UTF-8 convention.
+        // (Self-contained rather than reusing BindOutputColumnNonUtf16Unicode, which assumes a
+        // std::basic_string result and so does not cover SqlText / fixed-capacity narrow strings.)
+        if (cb.ServerType() == SqlServerType::POSTGRESQL)
+        {
+            auto u16String = std::make_shared<std::u16string>();
+            u16String->resize(StringTraits::Size(result) != 0 ? StringTraits::Size(result) : 255);
+
+            cb.PlanPostProcessOutputColumn([stmt, column, result, indicator, u16String]() {
+                if (*indicator == SQL_NULL_DATA)
+                    u16String->clear();
+                else if (*indicator == SQL_NO_TOTAL)
+                    ; // Length unknown; keep what the driver already wrote into the bound buffer.
+                else if (std::cmp_less_equal(*indicator, u16String->size() * sizeof(char16_t)))
+                    u16String->resize(static_cast<size_t>(*indicator) / sizeof(char16_t));
+                else
+                {
+                    // Truncation with a known total: grow and re-fetch the whole value.
+                    auto const totalChars = static_cast<size_t>(*indicator) / sizeof(char16_t);
+                    u16String->resize(totalChars + 1);
+                    auto const rv = SQLGetData(stmt,
+                                               column,
+                                               SQL_C_WCHAR,
+                                               u16String->data(),
+                                               static_cast<SQLLEN>((totalChars + 1) * sizeof(char16_t)),
+                                               indicator);
+                    (void) rv;
+                    assert(SQL_SUCCEEDED(rv));
+                    u16String->resize(totalChars);
+                }
+
+                auto const u8String = ToUtf8(*u16String);
+                // Resize may clamp to a fixed-capacity string's Capacity; copy only the post-clamp
+                // length so a UTF-8 expansion never overruns the inline buffer.
+                StringTraits::Resize(result, static_cast<SQLLEN>(u8String.size()));
+                std::memcpy(StringTraits::Data(result), u8String.data(), StringTraits::Size(result));
+
+                if constexpr (requires { StringTraits::PostProcessOutputColumn(result, *indicator); })
+                    if (*indicator != SQL_NULL_DATA && *indicator != SQL_NO_TOTAL)
+                    {
+                        auto const syntheticIndicator = static_cast<SQLLEN>(StringTraits::Size(result));
+                        StringTraits::PostProcessOutputColumn(result, syntheticIndicator);
+                    }
+            });
+
+            return SQLBindCol(stmt,
+                              column,
+                              SQL_C_WCHAR,
+                              static_cast<SQLPOINTER>(u16String->data()),
+                              static_cast<SQLLEN>(u16String->size() * sizeof(char16_t)),
+                              indicator);
+        }
+
         if constexpr (requires { AnsiStringType::Capacity; })
             StringTraits::Resize(result, AnsiStringType::Capacity);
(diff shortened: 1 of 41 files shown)


src/Lightweight/DataBinder/BasicStringBinder.hpp
```diff
@@ -39,6 +39,11 @@ namespace detail
     {
         using CharType = ArrayType::value_type;
 
+        // Character C types NUL-terminate every SQLGetData transfer, consuming the last buffer
+        // slot; SQL_C_BINARY transfers raw bytes WITHOUT a terminator, so every buffer slot holds
+        // data. Getting this wrong for binary loses the byte at each continuation boundary.
+        constexpr size_t NulChars = CType == SQL_C_BINARY ? 0 : 1;
```
christianparpart:
```suggestion
        constexpr size_t TrailingNulTerminators = CType == SQL_C_BINARY ? 0 : 1;
```
We should encode that this is about the number of trailing bytes that are to be included / cut off in a string/byte buffer, otherwise, the expressions below read really awkwards from a code maintainer's point of view.

src/Lightweight/DataBinder/UnicodeConverter.hpp
```diff
@@ -216,6 +216,18 @@ LIGHTWEIGHT_API std::u8string ToUtf8(std::u32string_view u32InputString);
 /// @ingroup Unicode
 LIGHTWEIGHT_API std::u8string ToUtf8(std::u16string_view u16InputString);
 
+/// Converts from UTF-16 to UTF-8, returned as a plain std::string holding the UTF-8 bytes.
+///
+/// Byte-identical to ToUtf8(); convenience for the many call sites that carry UTF-8 in
+/// std::string and would otherwise repeat the reinterpret_cast copy.
+///
+/// @ingroup Unicode
+[[nodiscard]] inline std::string ToUtf8String(std::u16string_view u16InputString)
```
christianparpart: seems like you don't use this function anywhere.
As discussed also in the past:

- UTF-32 strings should use std::u32string
- UTF-16 strings should use std::u16string
- UTF-8 strings should use std::u8strings
- system / codepage encoded strings should use std::string

`std::string ToUtf8String(std::u16string_view u16InputString)` explicitly violates this contract above. It's misleading.

And I also cannot find any user of this function, hence, dead code 😄 ➡️ 💀 kill it

src/Lightweight/SqlStatement.cpp
```diff
@@ -435,4 +489,376 @@ SqlQueryBuilder SqlStatement::QueryAs(std::string_view const& table, std::string
...
+        auto const utf8 = ToUtf8(std::u16string_view { cell, units });
+        return std::string { reinterpret_cast<char const*>(utf8.data()), utf8.size() };
+    }
+
+    char const* const cell = boundColumn.buffer.data() + (rowInBatch * boundColumn.elementWidth);
+    // The indicator carries the byte length of the value (excluding the NUL terminator). Clamp to
+    // the usable buffer width to defend against a driver over-reporting on SQL_SUCCESS_WITH_INFO.
+    auto const usableWidth = boundColumn.elementWidth - 1;
+    auto const length = indicator >= 0 ? std::min(static_cast<std::size_t>(indicator), usableWidth) : usableWidth;
+    return std::string { cell, length };
```
christianparpart: How do you deal with received clamped values here? We should deal with that, otherwise the backup would be corrupted.

src/Lightweight/SqlBackup/Backup.cpp
```diff
@@ -165,15 +263,233 @@ std::string BuildSelectQueryWithOffset(SqlQueryFormatter const& formatter,
...
+        // Text-family cells: GetString returns UTF-8 regardless of how the driver bound the column,
+        // matching the single-row path's std::string / (u16string -> ToUtf8) decode. Time reaches
+        // here only on PG/MSSQL (gated by TableIsArrayFetchable), where TIME is read as driver text.
+        return orNull(cursor.GetString(r, i), [](auto& v) -> BackupValue { return std::move(v); });
+    if (std::holds_alternative<Date>(colDef.type))
+        // Native SQL_DATE_STRUCT array bind; same std::format as the single-row SqlDate read.
+        return orNull(cursor.GetDate(r, i), [](auto const& v) -> BackupValue { return std::format("{}", v); });
+    if (std::holds_alternative<DateTime>(colDef.type) || std::holds_alternative<Timestamp>(colDef.type))
+        // Native SQL_TIMESTAMP_STRUCT array bind; same std::format as the single-row SqlDateTime read.
+        return orNull(cursor.GetTimestamp(r, i), [](auto const& v) -> BackupValue { return std::format("{}", v); });
```
christianparpart, line 434: A DATETIME SQL column on SQLite is read as a timestamp by the decoder, but bound as text by the driver, that looks like it'll end up in an type-mismatch exception.

src/Lightweight/SqlBackup/ChunkPlanner.cpp
```diff
@@ -0,0 +1,233 @@
...
+    constexpr size_t LobSizeThreshold = 1U << 20; // 1 MB: anything wider is treated as a LOB
+    for (auto const& c: table.columns)
+    {
+        bool const isLob = std::visit(
+            [](auto const& t) -> bool {
+                using T = std::decay_t<decltype(t)>;
+                if constexpr (std::is_same_v<T, Binary> || std::is_same_v<T, VarBinary> || std::is_same_v<T, Text>)
+                    return true;
+                else if constexpr (std::is_same_v<T, Varchar> || std::is_same_v<T, NVarchar>)
+                    return t.size == 0 || t.size >= LobSizeThreshold;
```
christianparpart: I don't quite get it. What exactly is with the values between 8k and 1 MB? We we have a test case covered for that?

src/Lightweight/SqlBackup/ChunkPlanner.cpp
```diff
@@ -0,0 +1,233 @@
...
+
+bool TableHasLobColumn(SqlSchema::Table const& table)
+{
+    // LOB = unbounded variable-length. The MSSQL batched reader represents varchar(max)/text as
+    // Varchar with the driver's max COLUMN_SIZE sentinel; treat very large declared sizes as LOB,
+    // and binary/varbinary/text variants as LOB regardless of size (they use SqlDynamicBinary today).
+    constexpr size_t LobSizeThreshold = 1U << 20; // 1 MB: anything wider is treated as a LOB
+    for (auto const& c: table.columns)
+    {
+        bool const isLob = std::visit(
```
christianparpart: The whole for-loop looks more like you would be interested in using something like this:

```cpp
auto const isLob = [](...) { ... };
return std::ranges::any_of(table.columns, isLob);
```

src/Lightweight/SqlBackup/ChunkPlanner.cpp
```diff
@@ -0,0 +1,233 @@
...
+                    return std::is_same_v<T, Integer> || std::is_same_v<T, Bigint> || std::is_same_v<T, Smallint>
+                           || std::is_same_v<T, Tinyint> || std::is_same_v<T, Real> || std::is_same_v<T, Bool>
+                           || std::is_same_v<T, Varchar> || std::is_same_v<T, Char> || std::is_same_v<T, Decimal>
+                           || std::is_same_v<T, NVarchar> || std::is_same_v<T, NChar> || std::is_same_v<T, Date>
+                           || std::is_same_v<T, DateTime> || std::is_same_v<T, Timestamp>;
+            },
+            c.type);
+        if (!safe)
+            return false;
+    }
```
christianparpart: The same applies like above:

```cpp
auto const isSafePred = [](...) { ... };
return std::ranges::any_of(table.columns, isSafePred) || !TableHasLobColumn(table);
```

src/Lightweight/SqlBackup/ChunkPlanner.cpp
```diff
@@ -0,0 +1,233 @@
...
+            continue;
+        bool const numeric = std::visit(
+            [](auto const& t) -> bool {
+                using T = std::decay_t<decltype(t)>;
+                return std::is_same_v<T, Integer> || std::is_same_v<T, Bigint> || std::is_same_v<T, Smallint>
+                       || std::is_same_v<T, Tinyint>;
+            },
+            c.type);
+        return numeric ? std::optional<std::string> { pkName } : std::nullopt;
+    }
```
christianparpart:
```cpp
auto const isNumericPred = [] (...) { ... };
if (std::ranges::any_of(table.columns | std::views::filter(...), isNumericPred);
    return std::optional<std::string> { pkName };
```
same rule like above.

src/Lightweight/SqlBackup/ChunkPlanner.cpp
```diff
@@ -0,0 +1,233 @@
...
+            // Wrap-safe span estimate (pkMax - pkMin in signed int64 is UB for the full range),
+            // saturating at SIZE_MAX for the degenerate full-span case.
+            auto const estimateMinus1 = static_cast<uint64_t>(pkMax) - static_cast<uint64_t>(pkMin);
+            state.totalRows.store(estimateMinus1 == std::numeric_limits<uint64_t>::max()
+                                      ? std::numeric_limits<size_t>::max()
+                                      : static_cast<size_t>(estimateMinus1) + 1);
+            auto windowIndex = uint32_t { 0 };
+            for (auto const& [lo, hi]: windows)
+            {
+                plan.chunks.push_back(Chunk {
```
christianparpart: why not emplace_back?

src/Lightweight/SqlBackup/ChunkPlanner.cpp
```diff
@@ -0,0 +1,233 @@
...
+            auto const windows = PlanPrimaryKeyWindows(pkMin, pkMax, width);
+            auto& state = plan.tableStates.emplace_back();
+            state.remainingChunks.store(windows.size());
+            // Wrap-safe span estimate (pkMax - pkMin in signed int64 is UB for the full range),
+            // saturating at SIZE_MAX for the degenerate full-span case.
+            auto const estimateMinus1 = static_cast<uint64_t>(pkMax) - static_cast<uint64_t>(pkMin);
+            state.totalRows.store(estimateMinus1 == std::numeric_limits<uint64_t>::max()
+                                      ? std::numeric_limits<size_t>::max()
+                                      : static_cast<size_t>(estimateMinus1) + 1);
+            auto windowIndex = uint32_t { 0 };
```
christianparpart: i wonder if we could get around this mutability by using `std::views::enumerate`.

src/Lightweight/SqlBackup/ChunkPlanner.cpp
```diff
@@ -0,0 +1,233 @@
...
+                    .arrayFetchable = arrayFetchable,
+                    .state = &state,
+                });
+            }
+        }
+        else
+        {
+            auto& state = plan.tableStates.emplace_back();
+            state.remainingChunks.store(1);
+            plan.chunks.push_back(Chunk {
```
christianparpart: why not `emplace_back`?

src/Lightweight/SqlBackup/ChunkPlanner.cpp
```diff
@@ -0,0 +1,233 @@
...
+                if constexpr (std::is_same_v<T, Time>)
+                    return timeIsText;
+                else if constexpr (std::is_same_v<T, Guid>)
+                    return guidIsNative;
+                else
+                    return std::is_same_v<T, Integer> || std::is_same_v<T, Bigint> || std::is_same_v<T, Smallint>
+                           || std::is_same_v<T, Tinyint> || std::is_same_v<T, Real> || std::is_same_v<T, Bool>
+                           || std::is_same_v<T, Varchar> || std::is_same_v<T, Char> || std::is_same_v<T, Decimal>
+                           || std::is_same_v<T, NVarchar> || std::is_same_v<T, NChar> || std::is_same_v<T, Date>
+                           || std::is_same_v<T, DateTime> || std::is_same_v<T, Timestamp>;
```
christianparpart: i know, there's no `std::is_any_of<T, U...>`, but we now have two places in the code to check that `T` satisfies any of the given other types.

src/tools/dbtool/BackupDiff.hpp
```diff
@@ -0,0 +1,33 @@
...
+///
+/// @param left  Path to the first ("left"/baseline) backup archive.
+/// @param right Path to the second ("right"/candidate) backup archive.
+/// @param ignoreTables Sanitized table names whose differences are reported but do NOT count
+///        as a failure. Use for high-write tables (e.g. log tables) that legitimately drift
+///        between two backups of a *live* database, so a stress comparison isn't drowned out
+///        by benign live-data churn. The match is on the data-path (sanitized) table name.
+/// @return EXIT_SUCCESS (0) if every common table (outside @p ignoreTables) has an identical
+///         row multiset and both archives contain the same set of tables; EXIT_FAILURE (1) if
+///         any non-ignored difference is found or either archive cannot be read.
```
christianparpart: at first i was scared that this is a duplicate of https://github.com/LASTRADA-Software/Lightweight/pull/484 - we should definitely try to avoid that.

Ok, this works on backup files instead. 🤔

I don't like the return type and the fact that you don't know where exactly the diff has happened (if any). You could add another parameter to optionally pass a DI-style streaming interface of diff events (if any). In case that's a non-cheap action, you could also send progress events through that.

NB: defaulted parameters are a hidden nightmare. Try to avoid that and enforce the caller to be always explicit. We really need a strong reason otherwise to disobey that. :)

christianparpart (changes requested):
overall okay, but a few things to mention / change.

src/tools/dbtool/BackupDiff.hpp
Yaraslaut: this functionality mostly used to make sure that the implementation is correct, not something user will actively use, maybe I should remove it?

src/Lightweight/SqlBackup/ChunkPlanner.cpp
Yaraslaut: added (claude) a comment there: 
```
    // LOB = unbounded variable-length. The MSSQL batched reader represents varchar(max)/text as
    // Varchar with the driver's max COLUMN_SIZE sentinel; treat very large declared sizes as LOB,
    // and binary/varbinary/text variants as LOB regardless of size (they use SqlDynamicBinary today).
 ```

src/Lightweight/SqlBackup/Backup.cpp
Yaraslaut: so, on my machine what i have is the following mapping:
`SQLDescribeCol reports DATE→SQL_TYPE_DATE and DATETIME→SQL_TYPE_TIMESTAMP natively, and the array path works` 
not as a text, so i am not sure what to do here, especially since we have a test that is covering this

src/Lightweight/SqlStatement.cpp
Yaraslaut: added a throw

src/tools/dbtool/BackupDiff.hpp
christianparpart: i think we can keep it, as a measure of quailty. :thinking:

PR #506, opened 2026-06-11
Add native batched INSERT/UPDATE for DataMapper

christianparpart:
Inserting or updating many records previously meant calling `Create`/`CreateExplicit`/`Update` in a loop, which re-prepares the statement for every single row. This adds a batched path that prepares once and submits the whole batch, preferring native ODBC row-wise array binding (a single zero-copy `SQLExecute`) and falling back to a prepare-once + per-row execute only when native binding cannot apply.

## Changes

- **`SqlStatement`**: new row-major `ExecuteBatch(rows, accessors...)` overload. Native row-wise binding (`SQL_ATTR_PARAM_BIND_TYPE = sizeof(row)`) is used when every bound column is row-bindable — primitives, `SqlDate`/`SqlTime`/`SqlDateTime`, `SqlNumeric`, or `std::optional` of a fixed non-numeric type — otherwise it falls back to a per-row soft path that handles every supported type (strings, binary, variant, …).
- **`std::optional` columns** stay zero-copy on the native path: the contained value is bound in place and per-row NULLs are supplied through a temporary, row-strided indicator buffer (new `SqlDataBinderCallback::ProvideBatchStagingBuffer`).
- **Eligibility** is data-driven via the opt-in `SqlIsNativeRowBindableValue` trait, and guarded by three gates: compile-time type/reference checks (incl. `sizeof(row) % alignof(SQLLEN) == 0` for optional columns), the new per-DBMS `SqlQueryFormatter::SupportsNativeRowBatch()` capability, and a runtime accessor-stride check.
- **`DataMapper`**: new `CreateAll` / `UpdateAll` accepting any contiguous, sized range of records (`std::vector`, `std::array`, `std::span`, C array; non-contiguous ranges are rejected at compile time). `UpdateAll` writes all storable non-primary-key columns matched on the primary key.
- **`SqlStatement::Prepare`** now resets the parameter-array binding attributes (`PARAMSET_SIZE` / `PARAM_BIND_TYPE` / `PARAM_BIND_OFFSET_PTR`) so a statement handle reused across batched and single executes — as `DataMapper` does — is never left in a stale row-wise binding state.
- `SqlDate`/`SqlTime`/`SqlDateTime` gain a column-wise `BatchInputParameter`, and the primitive batch binders accept an optional indicator pointer, to support the optional native path.

**Performance:** for fixed-width records the batch is a single prepare + single row-wise `SQLExecute` (zero-copy); for records with variable-length columns it is a single prepare + per-row execute — both eliminate the per-row re-prepare of the previous loop-based approach.

**Risk:** per-DBMS row-wise array binding behaviour is the main risk (gated by `SupportsNativeRowBatch()`); `SqlGuid` is deliberately excluded from native batching because SQLite binds it via a per-value text conversion. Covered by tests across SQLite, MS SQL Server and PostgreSQL.

### DIFF
--- a/src/Lightweight/DataBinder/Core.hpp
+++ b/src/Lightweight/DataBinder/Core.hpp
@@ -115,6 +115,13 @@ struct SqlBasicStringOperations;
 namespace detail
 {
 
+    /// @brief Satisfied when @p T is the same type as at least one of @p Us.
+    ///
+    /// Mirrors @c Lightweight::detail::OneOf, but lives in the DataBinder layer so the low-level binder
+    /// headers can use it without reaching up to the higher-level Utils.hpp.
+    template <typename T, typename... Us>
+    concept IsAnyOf = (std::same_as<T, Us> || ...);
+
     // clang-format off
 template <typename T>
 concept HasGetStringAndGetLength = requires(T const& t) {
--- a/src/Lightweight/DataBinder/Core.hpp
+++ b/src/Lightweight/DataBinder/Core.hpp
@@ -84,6 +84,10 @@ class LIGHTWEIGHT_API SqlDataBinderCallback
     ///
     /// @note The buffer contents are unspecified; the caller is responsible for initializing it.
     /// @note The returned storage is aligned to at least @c alignof(std::max_align_t).
+    /// @note Row-wise callers request @c ~rowStride*rowCount bytes to hold only @c rowCount @c SQLLEN
+    /// indicators. This over-allocation is intrinsic to ODBC row-wise binding, which strides the
+    /// @c StrLen_or_IndPtr array by @c SQL_ATTR_PARAM_BIND_TYPE (the row stride) — there is no separate
+    /// indicator stride — so a tightly packed indicator array would require descriptor-level binding.
     ///
     /// @param byteCount The number of bytes to provide.
     /// @return A pointer to the first byte of the buffer.
@@ -205,8 +209,11 @@ inline constexpr bool SqlIsNativeRowBindableValue = false;
 template <typename T>
 inline constexpr bool SqlIsNumericValue = false;
 
-/// @brief Whether @p T is a @c std::optional specialization. Distinct name from any DataMapper trait to
-/// avoid one-definition-rule clashes when both headers are included.
+/// @brief Whether @p T is a @c std::optional specialization.
+///
+/// Defined locally rather than reusing @c Lightweight::IsSpecializationOf (Utils.hpp) or the DataMapper
+/// @c IsStdOptional (Field.hpp): this low-level binder header is deliberately kept free of dependencies on
+/// those higher-level headers, so it carries its own minimal optional traits.
 template <typename T>
 inline constexpr bool SqlIsStdOptional = false;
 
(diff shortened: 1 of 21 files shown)


src/Lightweight/DataBinder/Core.hpp
```diff
@@ -169,6 +185,50 @@ concept SqlInputParameterBatchBinder =
...
+};
+
+template <typename T>
+struct SqlOptionalInner<std::optional<T>>
+{
+    using type = T; ///< The value type contained in the @c std::optional.
+};
+
+template <typename T>
+using SqlOptionalInnerType = typename SqlOptionalInner<T>::type;
```
Yaraslaut: optional has the `value_type` for this

src/Lightweight/DataBinder/Primitives.hpp
```diff
@@ -94,6 +107,28 @@ template <> struct SqlDataBinder<unsigned long long>: Int64DataBinderHelper<unsi
...
+template <> inline constexpr bool SqlIsNativeRowBindableValue<int8_t> = true;
+template <> inline constexpr bool SqlIsNativeRowBindableValue<uint8_t> = true;
+template <> inline constexpr bool SqlIsNativeRowBindableValue<int16_t> = true;
+template <> inline constexpr bool SqlIsNativeRowBindableValue<uint16_t> = true;
+template <> inline constexpr bool SqlIsNativeRowBindableValue<int32_t> = true;
+template <> inline constexpr bool SqlIsNativeRowBindableValue<uint32_t> = true;
+template <> inline constexpr bool SqlIsNativeRowBindableValue<int64_t> = true;
+template <> inline constexpr bool SqlIsNativeRowBindableValue<uint64_t> = true;
+template <> inline constexpr bool SqlIsNativeRowBindableValue<float> = true;
+template <> inline constexpr bool SqlIsNativeRowBindableValue<double> = true;
```
Yaraslaut: I think we can replace it with something like:
```
template<typename T>
inline constexpr bool SqlIsNativeRowBindableValue<T> = is_any_of<T, bool, char, ... , double>;
```

src/Lightweight/DataMapper/DataMapper.hpp
```diff
@@ -534,6 +578,14 @@ class DataMapper
...
             pkOverride = std::nullopt);
 
+    /// @brief Core implementation of CreateAll() operating on a contiguous span of records.
+    template <typename Record>
+    void CreateAllSpan(std::span<Record const> records);
+
+    /// @brief Core implementation of UpdateAll() operating on a contiguous span of records.
+    template <typename Record>
+    void UpdateAllSpan(std::span<Record const> records);
+
```
Yaraslaut: do not think that this should be separate api, nothing special in `span` for us here

src/Lightweight/DataMapper/DataMapper.hpp
```diff
@@ -1385,6 +1437,93 @@ RecordPrimaryKeyType<Record> DataMapper::CreateExplicit(Record const& record)
...
+    /// Accessor tuple for the WHERE clause of a batched UPDATE: primary-key columns.
+    template <std::size_t I, typename Record>
+    auto MakeUpdateWhereAccessor()
+    {
+        using FieldType = Reflection::MemberTypeOf<I, Record>;
+        if constexpr (IsPrimaryKey<FieldType>)
+            return std::tuple<FieldValueAccessor<I>> {};
+        else
+            return std::tuple<> {};
+    }
```
Yaraslaut, line 1492: hm, interesting utility

Yaraslaut (approved):
great addition, thanks :)

PR #507, opened 2026-06-12
Add C++23 coroutine support (async API)

christianparpart:
## Summary

Adds a **C++23 coroutine API** so you can drive Lightweight with coroutines — single- or multi-threaded — across SQLite, PostgreSQL and SQL Server on Windows, Linux and macOS. It is a **first-class, always-built part of the library** (no flag to enable).

True async ODBC isn't portable, so the model **offloads each blocking ODBC call to a worker thread and resumes your coroutine on a scheduler you inject** — your app thread is never blocked.

## Using it

Enable async on a connection once (you inject the executors), then `co_await`:

```cpp
DataMapper dm;
dm.Connection().EnableAsync(dbWorkers, appLoop);
```

Querying uses the **same fluent builder as the synchronous API** — start the chain with `QueryAsync<Record>()` instead of `Query<Record>()`, and every finisher (`All`, `First`, `First(n)`, `Range`, `Count`, `Exist`, `Delete`, and the field-projection variants) returns a `Task` of its usual result:

```cpp
auto active = co_await dm.QueryAsync<Person>()
                        .Where(FieldNameOf<&Person::is_active>, "=", true)
                        .OrderBy(FieldNameOf<&Person::name>, SqlResultOrdering::ASCENDING)
                        .First(10);                        // Task<std::vector<Person>>

auto total  = co_await dm.QueryAsync<Person>().Count();    // Task<std::size_t>
```

Record-level operations keep a dedicated `…Async` counterpart: `CreateAsync`, `QuerySingleAsync` (by primary key), `UpdateAsync`, `DeleteAsync`, `LoadRelationsAsync`. `Pool::AcquireAsync` acquires a pooled connection without blocking, and `AsyncSqlTransaction` scopes a transaction across `co_await`s.

**Lifetime:** async methods capture their operand (the record, or the query builder) by reference — keep the whole chain inside the `co_await` full-expression so it outlives the awaited task.

See `docs/async.md` for the full guide.

## What's included (`src/Lightweight/Async/`)

- `Task<T>` (lazy, exception-propagating) plus `SyncWait` / `SyncWaitPumping`.
- Injectable executors: `ThreadPool`, per-connection `Strand`, `Manual` (app run-loop), `Inline`.
- The offload core: `Async()`, `CancellationToken`, the `IAsyncBackend` seam and `ThreadOffloadBackend`.
- Async methods on the existing `SqlConnection` / `DataMapper` / `Pool` (no parallel type hierarchy), plus a distinct `AsyncSqlTransaction`.

The **synchronous public API and behavior are unchanged**, and there is **no new third-party dependency**.

## Testing

Async suite (28 cases / 87 assertions) validated against SQLite, mssql2022 and postgres; CI exercises the full matrix — SQLite, SQL Server 2017–2025, PostgreSQL; Windows/Linux/macOS; asan-ubsan + tsan; and the C++26 reflection build.

## Not included yet (designed, follow-up)

Native Windows + SQL Server event-based fast-path, low-level `SqlStatement` async members, and `Async::` module exports — they activate only on Windows + MSSQL and touch the core synchronous statement path, so they land where that combination can be validated.


docs/async.md
```diff
@@ -0,0 +1,230 @@
...
+#include <Lightweight/Async/ThreadPoolExecutor.hpp>
+#include <Lightweight/DataMapper/DataMapper.hpp>
+
+using namespace Lightweight;
+
+Async::ThreadPoolExecutor dbWorkers { 4 }; // 4 background DB threads
+Async::ManualExecutor     appLoop;         // your app/event-loop thread pumps this
+
+DataMapper dm;
+dm.Connection().EnableAsync(dbWorkers, appLoop);
```
Yaraslaut, line 55: I think we should add this into some global object, not in every connection, potentially similar to how we define default connection string, or in the data mapper pool, such that object you receive already support async if you defined it like this

docs/async.md
```diff
@@ -0,0 +1,230 @@
...
+
+// Back-pressure (suspend-on-exhaustion) requires the BoundedWait strategy; the default
+// DataMapperPool uses BoundedOverflow, which never suspends (see the note below).
+using BoundedPool = Pool<PoolConfig { .initialSize = 4, .maxSize = 8,
+                                      .growthStrategy = GrowthStrategy::BoundedWait }>;
+BoundedPool pool;
+
+Async::Task<std::optional<User>> Fetch(BoundedPool& pool, SqlGuid id)
+{
+    auto dm = co_await pool.AcquireAsync(dbWorkers, dbWorkers); // suspends when at capacity, never blocks
```
Yaraslaut: something is not right here, argument repeated two times

src/Lightweight/Async/CancellationToken.hpp
```diff
@@ -0,0 +1,71 @@
...
+/// token. Obtain a cancellable token via @ref Create; copies share the same cancellation
+/// state, so a caller can keep one copy and @ref Request cancellation on the operation
+/// holding another.
+///
+/// Cancellation is honored cooperatively and only @e before dispatch: the offload runtime checks
+/// @ref IsCancellationRequested before posting a step to the worker and, if set, completes it with
+/// @ref OperationCancelledError without ever occupying a worker. Once a step has begun running, the
+/// in-flight blocking ODBC call is @b not interrupted (there is no @c SQLCancel integration yet), so a
+/// request that arrives after dispatch only takes effect on the next not-yet-dispatched step.
+class CancellationToken
```
Yaraslaut: can we use `std::stop_token` instead?

src/Lightweight/Async/DataMapperAsync.hpp
Yaraslaut, line 1: this is very clean looking, great :)

src/Lightweight/Async/ManualExecutor.hpp
```diff
@@ -0,0 +1,83 @@
...
+            Work work;
+            if (!_queue.WaitAndPop(work, predicate))
+                return;
+            work();
+        }
+    }
+
+  private:
+    detail::WorkQueue _queue;
+    std::atomic<bool> _stopped { false }; ///< Set by Stop(); read by Run()'s wake condition.
```
Yaraslaut: here also, std::stop_token this is what we want to use most likely

src/Lightweight/Async/Task.hpp
Yaraslaut, line 1: I have a feeling that this overlaps too much with https://eel.is/c++draft/exec#task and i think we should use standard implementation when possible

src/Lightweight/CMakeLists.txt
```diff
@@ -140,6 +140,30 @@ set(SOURCE_FILES
...
+    Async/Fwd.hpp
+    Async/AsyncSqlTransaction.hpp
+    Async/DataMapperAsync.hpp
+)
+list(APPEND SOURCE_FILES
+    Async/ThreadPoolExecutor.cpp
+    Async/StrandExecutor.cpp
+    Async/ManualExecutor.cpp
+    Async/AsyncSqlTransaction.cpp
+)
```
Yaraslaut, line 165: this overlaps very much with what i am doing in the morph for the async execution as well, maybe we can find a middle ground for them so that we can create only one implementation and reuse it in both libraries

Yaraslaut (commented):
some comments, overall while this is a great addition i think we should separate async parts from the library itself and i would be much more in favor of something easier for the user to have, closer to the senders/receivers

docs/async.md
christianparpart: @Yaraslaut this might lead to unexpected results when you are not 100% in control of all queries being fired. Not sure we want that 🤔

christianparpart: i've added pool-wise configuration, so the connection (DM) comes pre-configured

docs/async.md
christianparpart: actually it was right, but i've adapted the code to avoid that alltogether.

src/Lightweight/CMakeLists.txt
christianparpart: Not too much. I've added an "Integrating with an external event loop / coroutine runtime" section documenting how to use IExecutor/IResumeScheduler adapter in such situations.

src/Lightweight/Async/Task.hpp
christianparpart: @Yaraslaut lol, this is C++26. not usable yet.

Yaraslaut: there is stdexec repo that provides implementation

Yaraslaut (approved):

PR #509, opened 2026-06-15
Introduce compile time optimizations

Yaraslaut:


christianparpart:
@Yaraslaut i really wanna get this in - before we release - when do you think you'll finish it? :D

PR #513, opened 2026-06-26
[SqlQuery] Remove unused SqlSelectQueryBuilder::Varying()

### LINKED ISSUE #495: Consider removal of  `SqlSelectQueryBuilder::Varying()`
I do not think that we use it, and the need for this functionality does not exist in my thinking, maybe we can drop this

Yaraslaut:
Closes #495.

## Summary
`Varying()` flipped the SELECT builder into a mode where the finalizers (`Count`/`All`/`First`/`Range`) returned a *copy* of the internal query instead of moving it out, so a single builder could be finalized into multiple query types. The default "Fluent" mode moves. Per #495 this feature is unused and unneeded, so it is removed.

## Changes
- **`SqlQuery/Select.hpp`** — removed the `SqlQueryBuilderMode` enum, the `Varying()` method, the `_mode` member, and the `Varying()` deducing-this override on `SqlSelectQueryStarter`; updated the gate doc comment to reference only `Distinct`.
- **`SqlQuery/Select.cpp`** — the four finalizers now unconditionally `std::move(_query)` (the former Fluent path; the `if (_mode == ...)` branches are gone).
- **`Lightweight.cppm`** — dropped the `SqlQueryBuilderMode` re-export.
- **`docs/sqlquery.md`** — removed `Varying()` from the starter documentation.
- **`tests/QueryBuilderTests.cpp`** — removed the now-obsolete "Varying" test case.

## Risk assessment
- **API**: breaking change for any external caller using `Varying()` / `SqlQueryBuilderMode` — appropriate for the issue's intent.
- **Behavior**: no change for existing default (Fluent) usage. Const finalizer overloads are unaffected (a truly-const builder always had Fluent mode, since `Varying()` was non-const).
- **Per-DBMS**: none — the change is purely about C++ move-vs-copy of the builder object and touches no DBMS-specific SQL generation.
- **Performance**: removes one per-finalize branch; negligible.

## Databases tested
- `sqlite3` — full suite green (1173 passed, 1 skipped).
- `mssql2022` (Docker, 16.00.4250) — full suite green (1171 passed, 3 skipped).
- `postgres` — **skipped**: no PostgreSQL ODBC driver available in this environment. Low risk given the change is database-agnostic.

Built with the `clang-debug` preset (PEDANTIC + ASan/UBSan + clang-tidy), no warnings.

### DIFF
--- a/docs/sqlquery.md
+++ b/docs/sqlquery.md
@@ -163,8 +163,8 @@ interface. Here we present a compressed list of functions that can be used to cr
 > the wildcard. `Fields("*")` and `Fields({"*"})` quote the literal and produce
 > `SELECT "*" FROM ...`, which is not what you want.
 >
-> `Distinct()` and `Varying()` are exposed on the starter as state-preserving
-> overrides (they return `SqlSelectQueryStarter&`), so chains like
+> `Distinct()` is exposed on the starter as a state-preserving override (it
+> returns `SqlSelectQueryStarter&`), so chains like
 > `Select().Distinct().Fields(...).All()` keep working while
 > `Select().Distinct().All()` is still a compile error. `Where`, `OrderBy`,
 > `GroupBy`, and the join family (`InnerJoin`, `LeftOuterJoin`, etc.) are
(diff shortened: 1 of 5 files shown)


christianparpart (approved):
just make sure that whatever you're removing, is not just a on-killing-spree-mode and make sure to adapt the internal code as needed (in case it's used)

Issue #515, opened 2026-06-26
HasMany resolves its inverse BelongsTo by field position, not by type/name


Yaraslaut:
## Summary

`HasMany<Other>` does not locate its inverse `BelongsTo` by matching the relationship type — it assumes the back-referencing `BelongsTo` in the *child* record sits at the **same field index** as the `HasMany` member in the *owner* record. When the indices don't line up, the generated `WHERE` clause filters on the wrong column and the relationship silently returns wrong/empty results (no error, no diagnostic).

## Where it happens

`ConfigureRelationAutoLoading` enumerates the owner record's members and, for a `HasMany`, captures the owner's member index `FieldIndex`, then builds the query with that same index against the *child* record:

- `src/Lightweight/DataMapper/DataMapper.hpp` — `ConfigureRelationAutoLoading` calls `BuildHasManySelectQuery<FieldIndex, ReferencedRecord>()` where `FieldIndex` is the index of the `HasMany` member in the **owner**.
- `BuildHasManySelectQuery` (same file) then emits:

```cpp
.Where(FieldNameAt<FieldIndex, OtherRecord>, SqlWildcard)
```

i.e. it takes the FK column name from `OtherRecord` (the child) **at the owner's HasMany index** — there is no search for the `BelongsTo` whose `ReferencedRecord` is the owner.

The same positional assumption is in `CallOnHasMany`.

## Why the existing tests don't catch it

The only direct `HasMany` coverage is `User` / `Email` (`src/tests/DataMapper/RelationTests.cpp`), which happens to align:

```cpp
struct User  { Field id; Field name;    HasMany<Email> emails; };  // emails at index 2
struct Email { Field id; Field address; BelongsTo<&User::id> user; }; // user   at index 2
```

Both relationship members land at index 2, so the positional shortcut produces the correct column. `Physician`/`Patient`/`Appointment` only exercise `HasManyThrough`, not direct `HasMany`.

## Minimal reproducer

```cpp
struct Employee;
struct Department {
    static constexpr std::string_view TableName = "Departments";
    Field<uint64_t, PrimaryKey::ServerSideAutoIncrement> id;        // 0
    Field<SqlAnsiString<40>> name;                                  // 1
    HasMany<Employee> employees;                                    // 2
};
struct Employee {
    static constexpr std::string_view TableName = "Employees";
    Field<uint64_t, PrimaryKey::ServerSideAutoIncrement> id;        // 0
[...]

PR #516, opened 2026-06-26
Add SQL-to-Lightweight documentation with test-backed examples

### LINKED ISSUE #368: Add example of SQL to Lightweight mapping in documentation
As discussed with @TooManyDots  one of the user friendly way for the documentation is to present examples on how some SQL queries are mapped to the Lightweight API, potentially using raw statements and DataMapper so that user can decide how they want to write it

Yaraslaut:
Closes #368.

## What

A new documentation page, `docs/sql-to-lightweight.md`, that presents how common SQL queries map to the Lightweight API across all three layers — raw `SqlStatement`, the query builder, and the high-level `DataMapper` — so users can pick the style that fits. Sections: SELECT (WHERE / AND-OR / IN / NULL / conditional filters / ORDER BY / LIMIT-TOP / pagination / DISTINCT / COUNT + aggregates / GROUP BY), JOINs (inner / left / multi-condition), INSERT (+ bulk), UPDATE, DELETE, relationships (`BelongsTo` / `HasMany`), CREATE TABLE / migrations, transactions, and custom result shapes.

## Examples are test-backed and kept in sync

Every C++ block in the doc is compiled and executed against a real database by `src/tests/DocExampleTests.cpp` (11 test cases, 74 assertions). Each block carries a `<!-- snippet: id -->` tag whose code is identical to the `//! [id]` region in the test. `scripts/check-doc-snippets.py` enforces that equality and fails CI on any drift (with a unified diff); it also has a `--fix` mode that regenerates the doc blocks from the tested regions. The check runs in the existing `check_docs` CI job, and the contract is documented in `AGENT.md`.

Workflow to change an example: edit the `//! [id]` region in the test, copy the lines into the matching doc block, run `python3 scripts/check-doc-snippets.py`.

## Testing

- **SQLite3** and **MS SQL Server 2022** (Docker) locally — all 74 assertions pass.
- **PostgreSQL** — covered by the CI `dbms_test_matrix`; no local ODBC driver was available to run it here.
- `clang-format` clean; the new test file is registered in `src/tests/CMakeLists.txt` so it runs in the full per-DBMS matrix.

## Notes

- While writing the tests I hit a `HasMany` resolution quirk (the inverse `BelongsTo` is matched by field position, not by type) — filed as #515. The doc/test schema keeps the relation members aligned and documents the constraint inline.
- No new dependencies; the sync checker is a standalone Python script.

## Risk

Low — additive. New doc page, one new test translation unit (additive to `LightweightTest`), one CI step, and registration lines. No library/runtime code changed.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -85,6 +85,8 @@ jobs:
         run: cmake --preset gcc-release -D LIGHTWEIGHT_BUILD_DOCUMENTATION=ON -D LIGHTWEIGHT_DOCS_WARN_AS_ERROR=ON
       - name: "check documentation"
         run: cmake --build --preset gcc-release --target doc
+      - name: "check doc code snippets are in sync with tested source"
+        run: python3 scripts/check-doc-snippets.py
 
   check_clang_tidy:
     name: "Check clang-tidy"
--- a/AGENT.md
+++ b/AGENT.md
@@ -209,6 +209,10 @@ If a database is **deliberately** skipped (e.g., feature genuinely doesn't apply
 
 When touching anything in `DataBinder/`, also exercise both `u8`/`u16`/`wchar_t` paths (the `WideChar` typedef in `Utils.hpp` selects the platform-appropriate UTF-16 code unit). See `.agent/testing.md` and `.agent/databases.md` for deeper guidance.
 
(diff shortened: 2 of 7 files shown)


christianparpart (approved):
❤️

PR #520, opened 2026-07-07
Combine code coverage across databases and ODBC driver versions

Yaraslaut:
## Problem

The Codecov number was effectively **SQLite3-only** and displayed as ~60% due to a metric artifact: (1) the coverage job started no database containers, so postgres/mssql runs failed to connect and were swallowed by `|| true`; (2) uploaded tracefiles carried branch records, and Codecov excludes every line with an untaken branch ("partial") from the headline.

## Result

**Codecov headline: 60% → 90.93%** (12,689/13,954 library lines, union of four database/driver legs, zero partials). The codecov/project check enforces a 90% floor.

## Changes

**`cmake/Coverage.cmake`** — strict per-environment targets `coverage-{sqlite3,postgres,mssql2022,mssql2017_odbc17}` (configurable via `COVERAGE_TEST_ENVIRONMENTS`): zero gcov counters, run the unit **and dbtool** suites against one database/driver combination, capture a filtered lcov tracefile. No branch records in uploads (true line coverage; branch stats remain in the local HTML target). `format` added to lcov ignore lists for Clang's line-0 coroutine-helper records (hard error in lcov ≥ 2.x).

**`.github/workflows/build.yml`** — coverage job installs PostgreSQL + MS SQL ODBC drivers (17 + 18), starts real `mssql2022`/`mssql2017`/`postgres` containers via the docker harness, runs the four per-env targets, uploads each tracefile under its own Codecov flag; Codecov merges flagged uploads into the combined headline.

**`codecov.yml`** — report scoped to the library (`src/Lightweight`; dbtool/ddl2cpp CLI plumbing excluded from the headline), patch status informational, project target 90%.

**Library fix found by the new tests** — `CharacterWidthFromDataType` didn't recognize PostgreSQL's `character varying`/`character` type names, so the lup-truncate width lookup resolved nothing on PostgreSQL and oversize INSERTs failed with `22001` instead of being clipped. Fixed in `SqlMigration.cpp` (PostgreSQL lengths are characters).

**~1,100 lines of new tests**, driven by per-file gap analysis of the merged tracefiles: select-builder overloads and const-delegation block; `ColumnType` for every type across all three dialects; AlterTable FK/index/IfExists dialect SQL incl. schema qualification; SQL Server `NCHAR` literal encoding (2/3/4-byte UTF-8, stray continuation); migration overlay (add/dedupe/clear/persist), plan folding (rename/drop side effects, FK commands, builder variants), post-connect default-schema hook, revert failure diagnostics (`RevertSingleMigration` + `RevertToMigration`), `RewriteChecksums`; live `INFORMATION_SCHEMA` width lookup; FK add/drop through the executor (SQLite table rebuild); CxxModelPrinter instantiation generation + report; SqlSchema composite-PK/multi-column-index introspection; lup-truncate wide-string/SqlText/UPDATE variants; SqlBackup PK-range single-row fallback for binary columns; `SqlVariant` DECIMAL scales 1–8; prefetch variant cursor over temporal/GUID columns; `HasOneThrough` auto-loading.

## Path to higher coverage (follow-up)

The remaining ~9% is predominantly driver-error/`catch` paths, transient-retry machinery in SqlBackup, and `SQL_NO_TOTAL` truncation loops — reachable only with an ODBC fault-injection layer. Realistic next milestone is 92% (target to be raised once reached); 99% would require that infrastructure or coverage-exclusion markers.

## Performance impact

None on the library at runtime (the one library change adds two string comparisons in a migration-time lookup). Coverage CI job runs 4 instrumented suite+dbtool legs (~+10 min).

## Risk assessment

- Library change is limited to the lup-truncate width-lookup type mapping; behavior on MSSQL/SQLite unchanged (verified), PostgreSQL now clips instead of failing.
- Strict per-env coverage targets fail the job on unreachable databases — intentional.
- One pre-existing state-dependent flake observed once on MSSQL (`SqlBackup: Backup and Restore` identity-table restore when `AllIdentityColumns` hits its swallow-exception path); passes in isolation and on rerun.

## Databases tested

- `sqlite3` (local): full suite + dbtool suite green (`coverage-sqlite3`).
- `mssql2022` (Docker, ODBC Driver 18, local): full suite + dbtool suite green (`coverage-mssql2022`).
- `postgres`, `mssql2017` + ODBC Driver 17 (CI): green in this PR's coverage job (the postgres leg caught the lup-truncate bug) and in `dbms_test_matrix`.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -951,7 +951,7 @@ jobs:
         # Rename the libsqliteodbc driver section to match the Windows driver
         # name so the shared .test-env.yml works on both platforms.
         run: |
-          sudo apt install -y cmake ninja-build catch2 unixodbc-dev sqlite3 libsqlite3-dev libsqliteodbc uuid-dev libyaml-cpp-dev libzip-dev lcov odbc-postgresql
+          sudo apt install -y cmake ninja-build catch2 unixodbc-dev sqlite3 libsqlite3-dev libsqliteodbc uuid-dev libyaml-cpp-dev libzip-dev lcov odbc-postgresql python3-yaml
           sudo sed -i 's/^\[SQLite3\]$/[SQLite3 ODBC Driver]/' /etc/odbcinst.ini
       - name: "Install MS SQL ODBC drivers (17 + 18)"
         # Both driver generations, so coverage includes the Driver 17 and
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -951,8 +951,18 @@ jobs:
         # Rename the libsqliteodbc driver section to match the Windows driver
         # name so the shared .test-env.yml works on both platforms.
         run: |
-          sudo apt install -y cmake ninja-build catch2 unixodbc-dev sqlite3 libsqlite3-dev libsqliteodbc uuid-dev libyaml-cpp-dev libzip-dev lcov
+          sudo apt install -y cmake ninja-build catch2 unixodbc-dev sqlite3 libsqlite3-dev libsqliteodbc uuid-dev libyaml-cpp-dev libzip-dev lcov odbc-postgresql
           sudo sed -i 's/^\[SQLite3\]$/[SQLite3 ODBC Driver]/' /etc/odbcinst.ini
+      - name: "Install MS SQL ODBC drivers (17 + 18)"
+        # Both driver generations, so coverage includes the Driver 17 and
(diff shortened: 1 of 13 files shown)


christianparpart (approved):
Interesting. Thanks, @Yaraslaut.
Also, did the Github UI just change the look? 🤔 
Fix all failures first, then shoot the shot :)

PR #529, opened 2026-07-28
Fix ddl2cpp duplicate includes and money/alias column mapping; bound SqlNumeric precision by what it can carry

### LINKED ISSUE #519: ddl2cpp: duplicate #include for repeated FK targets, and money columns emit precision that violates SqlNumeric's own cap
## Summary

Regenerating `src/entities/` in the Lastrada repo against the 6.10 schema (pinned Lightweight version `0.20260625.0`) surfaced two `ddl2cpp` regressions. Both broke the build; both were confirmed against the previous (correct) generation and, where relevant, against the live DB schema.

## 1. Duplicate `#include` lines when a table has multiple FKs to the same target

When a table has N foreign-key columns referencing the same target entity (e.g. several `*_FK` columns all pointing at the same lookup table), `ddl2cpp` emits `#include "Target.hpp"` once per FK column instead of deduplicating.

Confirmed as a regression: the previous generation had exactly one `#include` per referenced entity for the same tables.

**Example** (one entity header with 8 different FK columns to the same lookup table):

```cpp
#include "SomeOtherEntity.hpp"
#include "LookupTable.hpp"
#include "LookupTable.hpp"
#include "LookupTable.hpp"
#include "LookupTable.hpp"
#include "LookupTable.hpp"
#include "LookupTable.hpp"
#include "LookupTable.hpp"
#include "LookupTable.hpp"
#include "LookupTable.hpp"
#include "LookupTable.hpp"
#include "LookupTable.hpp"
#include "LookupTable.hpp"
#include "LookupTable.hpp"
#include "LookupTable.hpp"
#include "LookupTable.hpp"
#include "LookupTable.hpp"
#include "LookupTable.hpp"
```

**Scale:** ~98 generated files affected in one regeneration pass, ~400 duplicate lines total. Worst case: one entity header had a single target's `#include` repeated 55 times (one FK per text field referencing that target).

**Expected:** one `#include` per distinct referenced entity, regardless of how many FK columns reference it.

## 2. SQL Server `money` columns mapped to a precision that violates `SqlNumeric`'s own constraint

`ddl2cpp` maps `money` columns to `Light::SqlNumeric<19, 4>`. But `SqlNumeric` itself hard-caps precision:

```cpp
// DataBinder/SqlNumeric.hpp
static_assert(Precision <= SQL_MAX_NUMERIC_LEN); // SQL_MAX_NUMERIC_LEN == 16
```

So every entity with a `money` column fails to compile:

```
error: static assertion failed due to requirement 'Precision <= 16'
note: in instantiation of template class 'Lightweight::SqlNumeric<19, 4>' requested here
```

Verified these columns are genuinely `money` in the DB DDL and have never changed type (checked migration history back to the earliest `create table` for each) — this is not a real schema change, `ddl2cpp` is simply choosing a precision that its own emitted type can't hold. The previous generation mapped the same columns to `SqlNumeric<16, 4>`, which compiled fine.

**Scale:** 14 generated files / 30 fields, all `money`-typed price/cost/total columns across several unrelated tables.

**Expected:** either cap the emitted precision at 16 for `money` (matching `SQL_MAX_NUMERIC_LEN`), or raise/remove the `SqlNumeric` precision cap if 19 digits is intentionally supported by the ODBC binding — but the two must agree so generated code compiles.

## Possible third issue (lower confidence, only one instance found)

One `varchar(100)` column (confirmed via `sys.columns` against the live DB) has never changed type since the table was first created. It was correctly generated as `Light::SqlAnsiString<100>` previously, but came out as `Light::SqlWideString<50>` in this regeneration (wrong char width **and** wrong size). Only found this one instance while fixing consumer-code fallout, so it's possible this is isolated rather than systemic — flagging in case others hit the same thing on other columns.

## Environment

- Lightweight version: `0.20260625.0` (vcpkg port, `LASTRADA-Software/Lightweight` `v${VERSION}` tag)
- ddl2cpp/dbtool built from the same tag via the `tools` vcpkg feature
- Target DB: SQL Server (ODBC), ddl2cpp run against a live 6.10 schema instance

Yaraslaut:
Fixes #519, plus two `SqlNumeric` constraint bugs found while fixing it.

## 1. Duplicate `#include` for repeated FK targets

`CxxModelPrinter::TableInfo::requiredTables` was a `std::vector<std::string>` appended once **per FK column**, and `HeaderFileForTheTable()` sorted but never deduplicated — so a table with N foreign keys to the same target emitted that `#include` N times (the reporter saw one header repeat a single include 55 times, ~400 duplicate lines across 98 files).

Changed to `std::set<std::string>`, which dedupes and yields the same sorted order for free, so the explicit sort was dropped. Verified byte-identical Chinook golden output.

## 2. `money` mapped to a precision `SqlNumeric` rejected

`ddl2cpp` emits `SqlNumeric<19, 4>` for SQL Server `money`, but `SqlNumeric` asserted `Precision <= SQL_MAX_NUMERIC_LEN`, so every generated entity with a `money` column failed to compile.

`SQL_MAX_NUMERIC_LEN` is the size **in bytes** of `SQL_NUMERIC_STRUCT::val` (16 bytes = a 128-bit mantissa) — not a maximum decimal precision. The assert compared digits against bytes and rejected *every* decimal wider than 16 digits, including plain `decimal(18,2)`.

**The bound is now what this implementation can actually carry**, which is not the same as what the ODBC struct could hold. Two things narrow it:

- **The carrier.** `assign()` and `ToUnscaledValue()` fall back to `int64_t` when `__int128` is unavailable — and the guard excludes **both MSVC and clang-cl**. An `int64_t` unscaled value holds 18 digits; at `money`'s maximum the multiply exceeds `INT64_MAX`, the float→int conversion is UB, and x86-64 yields a sign-flipped `-922337203685477.5808`.
- **The readable width.** Every accessor except `ToUnscaledValue()` divides through `long double`, whose widest form (80-bit x87) carries 19 digits. Past that, nothing is readable on any platform.

So: **19 with a 128-bit carrier, 18 without.** `decimal(18, s)` compiles everywhere and `money` compiles wherever `__int128` exists; wider columns must be read as strings. `docs/data-binder.md` now tabulates what each accessor delivers — including that `operator==` compares via `ToFloat()` (7 digits).

## 3. `static_assert(Scale < Precision)` rejected legal types

`decimal(4,4)` — all four digits after the point, values in [0,1) — is legal in the SQL standard and in all three backends. Traced every conversion path first (`assign`, `ToUnscaledValue`, `ToFloat/Double/LongDouble`, `ToString`): the value is stored as the unscaled integer `value × 10^Scale` and nothing assumes an integral digit. Relaxed to `Scale <= Precision`, with `static_assert(Precision > 0)` added to preserve the rejection of `SqlNumeric<0,0>` the strict `<` gave for free.

## 4. MSSQL alias types resolved to the wrong column type

The batched `sys.columns` reader joined `sys.types` on `user_type_id` only, so alias types (`CREATE TYPE X FROM nvarchar(50)`) and `sysname` matched nothing and fell through to a `Varchar{max_length}` default — an `nvarchar(50)` became `SqlAnsiString<100>`: wrong char width *and* bytes-instead-of-chars size.

Fixed with `COALESCE(bt.name, ty.name)` + `LEFT JOIN sys.types bt ON ty.system_type_id = bt.user_type_id`. Verified against a live container that `user_type_id` has no duplicates (no row multiplication) and that no ordinary column's resolved type changes.

This is the same symptom class as the reporter's third, low-confidence item (a `varchar(100)` generated as `SqlWideString<50>`). That exact case **could not be reproduced** — it needs the force-unicode config *and* `Varchar{50}` — and is not claimed as fixed.

## Risk assessment

- The precision bound is *raised* relative to `master` (16 → 18/19), so no previously-compiling record stops compiling. `SqlNumeric<0,0>` was already rejected.
- Above 15 digits, `SqlNumeric` is exact only on the native `SQL_C_NUMERIC` path, which is unavailable on SQLite and MS SQL Server (`NativeNumericSupportIsBroken`) *and* on the Windows psqlODBC build. That is pre-existing, now documented, and worth its own issue — `NativeNumericSupportIsBroken()` keys off `SqlServerType`, but losslessness is a property of the driver build.
- The `sys.types` join is MSSQL-only and additive.

## Known issues confirmed but not fixed here

- `ToUnscaledValue()` reads `sqlValue.val` via `reinterpret_cast<__int128 const*>` at offset 3 of a 1-aligned struct — a misaligned load, tolerated on x86-64.
- `OutputColumn` passes `sizeof(ValueType)` as the `SQL_C_NUMERIC` buffer length for a 19-byte target followed immediately by `nativeValue`.
- `SqlSchema.cpp` now stores the resolved base-type name in `Column::dialectDependantTypeString`, so a declared alias name is no longer recoverable. Inert today — no consumer in `src/`.

## Databases tested

`sqlite3` (3.49.1), `mssql2022` (16.00.4250, Docker), `postgres` (16.4, Docker) — full suite, 0 failures on each.

## Compilers tested

`clang-debug` (clang-tidy + ASan/UBSan, `-Werror`), `gcc-release` (the compiler CI's database legs actually run), `gcc-release -D LIGHTWEIGHT_BUILD_MODULES=ON` (the module build, which rejects an exported entity with internal linkage — hence `inline constexpr`), and the Doxygen docs build with `LIGHTWEIGHT_DOCS_WARN_AS_ERROR=ON`.

### DIFF
--- a/src/Lightweight/DataBinder/SqlNumeric.hpp
+++ b/src/Lightweight/DataBinder/SqlNumeric.hpp
@@ -26,18 +26,61 @@ namespace Lightweight
 #endif
 // clang-format on
 
-/// Number of decimal digits that fit into the mantissa of an ODBC `SQL_NUMERIC_STRUCT`.
+namespace detail
+{
+
+    /// Number of whole decimal digits that fit into a binary magnitude of `bits` bits.
+    ///
+    /// `floor(bits * log10(2))`, evaluated with integer arithmetic (`log10(2) ~= 0.30103`) so that
+    /// it stays usable in a constant expression without pulling in `<cmath>` at compile time.
+    ///
+    /// @param bits Width of the binary magnitude, excluding any sign bit.
+    /// @return Number of decimal digits that magnitude represents without loss.
+    [[nodiscard]] constexpr std::size_t DecimalDigitsForBits(std::size_t bits) noexcept
+    {
+        return (bits * 30103) / 100'000;
+    }
+
+} // namespace detail
+
+/// Widest `Precision` a `SqlNumeric` may declare, in decimal digits.
 ///
-/// `SQL_MAX_NUMERIC_LEN` is the size *in bytes* of `SQL_NUMERIC_STRUCT::val` (16 bytes, i.e. a
-/// 128-bit unsigned mantissa) — it is not a maximum decimal precision. The number of decimal
-/// digits that can be stored is `floor(bits * log10(2))`, computed here with integer arithmetic
-/// (`log10(2) ~= 0.30103`) so it stays usable in a constant expression: 128 bits -> 38 digits,
-/// which is also the maximum `DECIMAL`/`NUMERIC` precision of MS SQL Server and PostgreSQL.
+/// This is deliberately *not* derived from `SQL_MAX_NUMERIC_LEN`. That macro is the size **in
+/// bytes** of `SQL_NUMERIC_STRUCT::val` (16, i.e. a 128-bit mantissa), not a decimal precision, so
+/// the historical `static_assert(Precision <= SQL_MAX_NUMERIC_LEN)` was a category error that
+/// rejected perfectly ordinary columns such as `DECIMAL(18, 2)`.
+///
+/// Bounding by the mantissa's theoretical capacity instead (128 bits -> 38 digits) would be the
+/// same category error in the other direction: 38 is what the ODBC *struct* could hold, not what
+/// this implementation can carry. Two things narrow it:
+///
+/// - **The unscaled carrier.** `assign()` and `ToUnscaledValue()` keep `value * 10^Scale` in a
+///   `LIGHTWEIGHT_INT128_T` where the toolchain offers a 128-bit integer, and in an `int64_t`
+///   otherwise — MSVC and clang-cl both take the latter path. A signed 64-bit carrier holds
+///   `DecimalDigitsForBits(63)` == 18 digits. At `Precision == 19` the conversion overflows:
+///   MS SQL Server's `money` maximum 922337203685477.5807 has the unscaled value
+///   9223372036854775808, one past `INT64_MAX`, so the `static_cast<int64_t>` of that
+///   floating-point value is an out-of-range conversion — undefined behaviour, which on x86-64
+///   yields `INT64_MIN` and renders as -922337203685477.5808, sign flipped.
+/// - **The readable width.** Every accessor except `ToUnscaledValue()` — `ToFloat`, `ToDouble`,
+///   `ToLongDouble` and therefore `ToString` — divides through `long double`, so no more digits can
+///   be read back than that type's significand holds. On the 80-bit x87 `long double` that is
+///   `DecimalDigitsForBits(64)` == 19; a fetched `SqlNumeric<20, 0>` holding 99999999999999999999
+///   already prints as 100000000000000000000.
+///
+/// Hence 18 without a 128-bit carrier and 19 with one. `DECIMAL(18, s)` compiles everywhere and
+/// `money` (`DECIMAL(19, 4)`) compiles wherever `__int128` exists; a wider column must be read as a
+/// string. `docs/data-binder.md` tabulates what each accessor delivers below that bound.
 ///
 /// NB: `inline` is load-bearing. At namespace scope `constexpr` implies `const`, hence internal
 /// linkage, and an exported template in the module interface (`SqlNumeric`, via its static_assert)
 /// may not reference an internal-linkage entity.
-inline constexpr std::size_t SqlMaxNumericPrecision = (std::size_t { SQL_MAX_NUMERIC_LEN } * 8 * 30103) / 100'000;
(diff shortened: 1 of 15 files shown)


src/Lightweight/DataBinder/SqlNumeric.hpp
```diff
@@ -26,12 +26,79 @@ namespace Lightweight
...
+///
+/// NB: `inline` is load-bearing. At namespace scope `constexpr` implies `const`, hence internal
+/// linkage, and an exported template in the module interface (`SqlNumeric`, via its static_assert)
+/// may not reference an internal-linkage entity.
+inline constexpr std::size_t SqlMaxNumericPrecision =
+#if defined(LIGHTWEIGHT_INT128_T)
+    detail::DecimalDigitsForBits(64); // 19 — bounded by the `long double` significand every accessor divides through
+#else
+    detail::DecimalDigitsForBits(63); // 18 — bounded by the `int64_t` magnitude that carries the unscaled value
+#endif
```
Yaraslaut: I do not like the idea that this is not static, since database should be the same and not depent on which platform you compiled executable

PR #541, opened 2026-08-08
test(coverage): cover SqlVariant type branches and the backup retry policy

Yaraslaut:
Follow-up to #539, which made never-instantiated DataMapper templates *visible* to coverage. This pass targets the next-largest gaps that are reachable without a live fault.

## What this adds

**1. `SqlVariant` type branches** (`SqlVariantDbTests.cpp`)

`SqlVariant::GetColumn` dispatches on the ODBC concise type. Two arms had no coverage on any database in the matrix:

- `SQL_TINYINT` — no test in the suite created a TINYINT column at all.
- The DECIMAL/NUMERIC scale switch's `default:` arm (every scale >= 9). The existing "by scale" test enumerates 0..8 only, so the fallthrough was never taken.

Both follow the conventions already in that file: the DECIMAL test carries the same `UNSUPPORTED_DATABASE(SQLITE)` guard (no native DECIMAL) and the same `MICROSOFT_SQL` value-assertion guard as its neighbour, since the MSSQL driver returns `SQL_C_NUMERIC` at scale 0 unless `SQL_DESC_SCALE` is set on the IRD. The switch arm still executes on every DBMS; only the round-tripped value is asserted where it is reliable.

**2. The backup retry policy** (`SqlBackup/Common.{hpp,cpp}`, `CommonHelpersTests.cpp`)

The backup/restore workers funnel every ODBC call through two retry loops, but neither had any test — reaching the retry arm meant provoking a real transient driver failure mid-backup. Together they account for a large share of the uncovered lines in `Backup.cpp` (134) and `Restore.cpp` (114).

Extracts the decision, mirroring `ClassifyOdbcResult` from 7d5e21c0:

```cpp
ClassifyRetryOutcome(SqlErrorInfo const&, unsigned attemptsSoFar, RetrySettings const&) -> RetryAction
```

It performs no I/O and touches no handle, so a test drives it by constructing a `SqlErrorInfo`. Both loops (`ConnectWithRetry` and the `RetryOnTransientError` template) now delegate to it, so the "is it transient / is the budget spent" rule lives in one place instead of two duplicated conditions.

`RetryOnTransientError` is also covered end-to-end: being a template over an arbitrary callable, a lambda throwing scripted `SqlException`s reaches every branch without a database — immediate success, transient-then-success, non-transient rethrow, budget exhaustion, and a void-returning instantiation.

## Risk

Behaviour is unchanged. The extracted predicate is the original condition verbatim and the success path is untouched; the only production edit is routing two existing conditions through one function. `RetryAction` is a new public enum in the `SqlBackup::detail` namespace, and `<cstdint>` was added to keep the header self-contained.

**Verified to have teeth by mutation**: relaxing the budget boundary from `>=` to `>` fails 3 cases / 4 assertions, including the end-to-end retry-count check.

## Not attempted

The remaining uncovered lines in `Backup.cpp`/`Restore.cpp` are the reconnect and partial-flush bodies *inside* those loops. Reaching them needs a connection-level fake (`SqlConnection::Connect`/`LastError` are non-virtual on a concrete class), which is a larger design change than this pass.

## Testing

- **Databases**: sqlite3, mssql2022 (Docker), postgres 16.4 (Docker) — all exit 0, 1327 cases each (up from 1316).
- **Compiler**: clangcl-release (Windows), clean under `/W4 /permissive-`.
- **Not verified locally**: GCC, the modules leg, the C++26 reflection leg, and the coverage delta (the coverage presets are Linux-only). Note `Common.hpp` gained a namespace-scope `enum` — the modules leg is worth watching.

New tests: 11 cases / ~50 assertions, 9 of which need no database.


christianparpart:
@Yaraslaut now needs a rebase 😇

christianparpart (approved):
Thanks, @Yaraslaut 😄

PR #546, opened 2026-08-13
fix: resolve issues #535, #536, #542, #543, #544, #545

### LINKED ISSUE #535: std::formatter<SqlDynamicBinary<N>> fails to compile (no begin()/end())
## Summary

`std::formatter<Lightweight::SqlDynamicBinary<N>>` (in `Lightweight/DataBinder/SqlDynamicBinary.hpp`) fails to compile because it iterates the type with a range-`for` loop, but `SqlDynamicBinary<N>` has no `begin()`/`end()` — it only exposes `data()`, `size()`, `resize()`, `empty()`, `clear()`.

```
error C3312: no callable 'begin' function found for type 'const Lightweight::SqlDynamicBinary<2147483647>'
```

## Where

```cpp
template <std::size_t N>
struct std::formatter<Lightweight::SqlDynamicBinary<N>>: std::formatter<std::string>
{
    auto format(Lightweight::SqlDynamicBinary<N> const& text, format_context& ctx) const
    {
        std::string humanReadableText;
        for (auto byte: text)   // <-- SqlDynamicBinary has no begin()/end()
        {
            ...
        }
        ...
    }
};
```

## How it's reached

`Lightweight::Field<T, ...>` has its own formatter that forwards to `std::formatter<T>`:

```cpp
template <typename T, auto P1, auto P2>
struct std::formatter<Lightweight::Field<T, P1, P2>>: std::formatter<T> { ... };
```

So any entity declaring a field like:

```cpp
Light::Field<std::optional<Light::SqlDynamicBinary<2147483647>>, Light::SqlRealName{"CONTENT"}> content;
```

pulls in `std::formatter<SqlDynamicBinary<N>>` as soon as something needs that field's formatter to be well-formed (e.g. via `DataMapper::Inspect`, or a `std::formattable`/similar trait check on the field type during `DataMapper::Create`/`Update`). This fails to compile with MSVC (cl.exe, C3312) the moment it's actually instantiated.

## Minimal repro

```cpp
#include <Lightweight/DataBinder/SqlDynamicBinary.hpp>
#include <format>

int main()
{
    Lightweight::SqlDynamicBinary<255> blob;
    std::format("{}", blob);   // C3312: no callable 'begin' function
}
```

## Expected

`SqlDynamicBinary<N>` should either:
- gain `begin()`/`end()` (e.g. delegating to the underlying buffer), or
- have its `std::formatter` reimplemented to iterate via `data()`/`size()` instead of a range-`for`, matching the existing `SqlDataBinder<SqlDynamicBinary<N>>::Inspect()` static method a few lines above, which already does this correctly:
  ```cpp
  static LIGHTWEIGHT_FORCE_INLINE std::string Inspect(ValueType const& value)
  {
      return std::format("SqlBinary<{}>(size={})", N, value.size());
  }
  ```

## Environment

- Lightweight `v0.20260625.0` (vcpkg)
- MSVC (cl.exe), C++23
- Windows x64

## Impact

Blocks any use of `DataMapper` CRUD (`Create`/`Update`/`QuerySingle`) against an entity that has a `varbinary(max)`-style column typed as `Light::Field<Light::SqlDynamicBinary<N>, ...>` or `Light::Field<std::optional<Light::SqlDynamicBinary<N>>, ...>` — the field declaration alone is fine, but any code path that needs the field's formatter fails to compile.

### LINKED ISSUE #536: DataMapper::Create narrows size_t to int for ServerSideAutoIncrement PK (C2397 under /WX)
## Summary

`DataMapper::CreateInternal` fails to compile for any record whose `ServerSideAutoIncrement` primary key field has a narrower type than `size_t` (e.g. `int32_t`/`int`, the common case for a SQL Server `int IDENTITY` column) under MSVC with `/W3 /WX` (warnings-as-errors, a common project configuration).

```
error C2397: conversion from 'size_t' to 'int' requires a narrowing conversion
```

## Where

`Lightweight/DataMapper/DataMapper.hpp`, in `DataMapper::CreateInternal`:

```cpp
if constexpr (HasAutoIncrementPrimaryKey<Record>)
    return { _stmt.LastInsertId(RecordTableName<Record>) };   // <-- narrows size_t -> int here
```

`SqlStatement::LastInsertId` is declared:

```cpp
[[nodiscard]] LIGHTWEIGHT_API size_t LastInsertId(std::string_view tableName);
```

`RecordPrimaryKeyType<Record>` for a record whose primary key field is `Field<int32_t, PrimaryKey::ServerSideAutoIncrement, ...>` resolves to `int`. The brace-init `return { size_t_value }` into an `int`-returning function is a narrowing conversion, which is ill-formed in list-initialization and triggers MSVC C2397 (and would trigger `-Wnarrowing`/similar warnings-as-errors on other compilers too) whenever the caller's project treats warnings as errors.

## Minimal repro

```cpp
#include <Lightweight/DataMapper/DataMapper.hpp>

struct Widget
{
    static constexpr std::string_view TableName = "WIDGET";
    Light::Field<int32_t, Light::PrimaryKey::ServerSideAutoIncrement, Light::SqlRealName{"ID"}> id;
    Light::Field<int32_t, Light::SqlRealName{"VALUE"}> value;
};

template <>
struct Lightweight::Description<Widget>
{
    static constexpr std::size_t FieldCount = 2;
    using Members = Lightweight::RecordMemberList<&Widget::id, &Widget::value>;
    static constexpr std::array<std::string_view, 2> FieldNames = { "ID", "VALUE" };
};

void insert(Lightweight::DataMapper& dm)
{
    Widget w;
    w.value = 42;
    dm.Create(w);   // C2397 under /W3 /WX (MSVC) -- narrows size_t to int in DataMapper.hpp
}
```

## Expected

`CreateInternal` should cast explicitly rather than rely on implicit brace-init narrowing, e.g.:

```cpp
if constexpr (HasAutoIncrementPrimaryKey<Record>)
    return static_cast<RecordPrimaryKeyType<Record>>(_stmt.LastInsertId(RecordTableName<Record>));
```

so a narrower (but still large enough) primary key type is a deliberate, silent truncation the caller opted into by choosing that column type, rather than a hard compile error under warnings-as-errors.

## Environment

- Lightweight `v0.20260625.0` (vcpkg)
- MSVC 19.51 (Visual Studio 18), `/W3 /WX`, C++23 (`-std:c++latest`)
- Windows x64

## Impact

Blocks `DataMapper::Create` on any record with a `ServerSideAutoIncrement` primary key typed narrower than `size_t` -- i.e. essentially every SQL Server `int IDENTITY` column mapped to `int32_t`, which is the standard/expected mapping -- when the consuming project builds with warnings-as-errors. Related to, but independent of, #535 (the `SqlDynamicBinary<N>` formatter issue) -- both surfaced while adding the first `DataMapper::Create` call in our codebase against a genuinely `IDENTITY`-backed `int32_t` primary key.

### LINKED ISSUE #542: SqlStatement.hpp includes <Windows.h> without NOMINMAX, breaking std::min/max in consumer TUs
## Summary

`SqlStatement.hpp` includes `<Windows.h>` on Windows without `NOMINMAX` (or `WIN32_LEAN_AND_MEAN`), so `Windows.h`'s `min`/`max` function-like macros leak into every translation unit that includes this header (directly or transitively). Any later use of `std::min`, `std::max`, `std::numeric_limits<T>::min()`/`max()`, or any user type's own `.min()`/`.max()` method in the *same TU* fails to compile, because the preprocessor rewrites the token `min`/`max` before the compiler ever sees it as an identifier.

## Where

```cpp
// src/Lightweight/SqlStatement.hpp
#if defined(_WIN32) || defined(_WIN64)
    #include <Windows.h>
#endif
```

No `#define NOMINMAX` precedes this include, and no `#undef min` / `#undef max` follows it.

This also causes `SqlStatement.hpp` itself to fail to compile in its own `FetchAllRowWise`:

```cpp
// src/Lightweight/SqlStatement.hpp (around line 1554, v0.20260625.0)
auto const minDepth = std::min(RowArrayCursor::MinArrayDepth, arrayDepth);
```

```
error C2589: '(': illegal token on right side of '::'
note: This diagnostic occurred in the compiler generated function
      'void Lightweight::SqlStatement::FetchAllRowWise(...)'
```

## Repro

- MSVC (cl.exe), Windows 11, C++23 (`/std:c++latest`), Lightweight `v0.20260625.0`.
- Consume `Lightweight::SqlStatement` (e.g. via `DataMapper`) from any project that does **not** itself define `NOMINMAX` before pulling in Lightweight's headers.
- Any consumer TU that includes a Lightweight header and later calls `std::min`/`std::max`/`std::numeric_limits<T>::min()` fails with `C2589`/`C4003`-style errors, because the macro substitution happens in *their* TU too, not just inside Lightweight's own headers.

Example from a downstream consumer (unrelated header, same TU as a Lightweight include, minimal repro of the blast radius):

```cpp
inline constexpr std::int64_t kOverrideDisabled = std::numeric_limits<std::int64_t>::min();
```

```
error C4003: not enough arguments for function-like macro invocation 'min'
error C2589: '(': illegal token on right side of '::'
error C2059: syntax error: ')'
```

## Suggested fix

Guard the `<Windows.h>` include in `SqlStatement.hpp` (and any other header that includes it, e.g. `DataBinder/Core.hpp`, `SqlConnection.hpp`, `SqlError.hpp`, `SqlTransaction.hpp`, `SqlColumnTypeDefinitions.hpp`) with `NOMINMAX`/`WIN32_LEAN_AND_MEAN`, e.g.:

```cpp
#if defined(_WIN32) || defined(_WIN64)
    #ifndef NOMINMAX
        #define NOMINMAX
    #endif
    #ifndef WIN32_LEAN_AND_MEAN
        #define WIN32_LEAN_AND_MEAN
    #endif
    #include <Windows.h>
#endif
```

This is the standard, non-breaking way for a header-only/library-included `<Windows.h>` to avoid clobbering every consumer's TU-wide namespace. Today, every consumer must remember to globally define `NOMINMAX` before including any Lightweight header, which is easy to miss (we only found it after a fresh full-build failure) and isn't documented anywhere in the README/AGENT.md.

## Workaround used downstream

We added a project-wide `add_compile_definitions(NOMINMAX WIN32_LEAN_AND_MEAN)` (guarded by `WIN32`) ahead of pulling in Lightweight via `FetchContent`, which resolves it for us — but it would be better fixed at the source so every consumer doesn't need to know about it.

### LINKED ISSUE #543: Windows build hard-requires libzip via find_package(CONFIG REQUIRED) with no CPM fallback or docs
## Summary

On Windows, `src/Lightweight/CMakeLists.txt` unconditionally requires `find_package(libzip CONFIG REQUIRED)` (a hard `REQUIRED`, no fallback), but this dependency is undocumented in the README/AGENT.md's build instructions and isn't declared in any vcpkg manifest shipped in the repo. A consumer that fetches Lightweight via `FetchContent` and provides `yaml-cpp`/`Catch2` (the two dependencies the docs do mention) still fails to configure, because `libzip` is missing too.

## Where

```cpp
// src/Lightweight/CMakeLists.txt
if(WIN32)
    find_package(libzip CONFIG REQUIRED)
    # vcpkg provides libzip::zip, create alias to match CPM target name
    add_library(zip ALIAS libzip::zip)
endif()
# Note: libzip for non-Windows is handled in root CMakeLists.txt (system pkg-config or CPM fallback)
```

Non-Windows platforms get a graceful pkg-config-or-CPM fallback (root `CMakeLists.txt`, "libzip support (Linux only)" section) — Windows gets neither fallback nor documentation, only a hard `REQUIRED` `find_package`.

## Repro

```
CMake Error at .../vcpkg/scripts/buildsystems/vcpkg.cmake:910 (_find_package):
  Could not find a package configuration file provided by "libzip" with any
  of the following names:
    libzip.cps
    libzipConfig.cmake
    libzip-config.cmake
Call Stack (most recent call first):
  src/Lightweight/CMakeLists.txt:22 (find_package)
```

Happens on any Windows configure (vcpkg-toolchain or not) where the consumer's vcpkg manifest doesn't already happen to include `libzip`.

## Suggested fix

Either:
1. Apply the same CPM fallback Windows already gets for `reflection-cpp`/`stdexec` (`find_package(libzip CONFIG QUIET)`, then `CPMAddPackage(... nih-at/libzip ...)` if not found) — the same source tree already does this for `libzip` on non-Windows, so the logic exists, just gated to `NOT WIN32`; or
2. If a system/vcpkg `libzip` is intentionally required on Windows, document it explicitly (README build prerequisites, `AGENT.md`) alongside `yaml-cpp` and `Catch2`, and/or ship a vcpkg manifest snippet consumers can copy.

## Workaround used downstream

Added `libzip` to our own `vcpkg.json` dependencies list, alongside `yaml-cpp` (see also #542 for a related undocumented-dependency issue).

### LINKED ISSUE #544: Static build (LIGHTWEIGHT_BUILD_SHARED=OFF) fails at generate time: $<TARGET_PDB_FILE:Lightweight> invalid for non-linker artifacts
## Summary

`src/Lightweight/CMakeLists.txt`'s `install()` rules unconditionally reference `$<TARGET_PDB_FILE:Lightweight>` on `WIN32`, regardless of whether `Lightweight` was configured as a static or shared library (`LIGHTWEIGHT_BUILD_SHARED`). `TARGET_PDB_FILE` is only valid for targets with a linker-created artifact (a DLL or EXE) — a static `.lib` has no such PDB, and CMake fails at **generate** time (not even at `cmake --install` time) as soon as `LIGHTWEIGHT_BUILD_SHARED=OFF` is set on Windows.

## Where

```cpp
// src/Lightweight/CMakeLists.txt
# Only install the PDB file for Debug and RelWithDebInfo configurations on Windows platform
if(WIN32)
    install(FILES $<TARGET_PDB_FILE:Lightweight>
        DESTINATION ${CMAKE_INSTALL_BINDIR}
        CONFIGURATIONS Debug RelWithDebInfo
        COMPONENT DebugSymbols
        OPTIONAL
    )
endif()
```

## Repro

```
cmake -S . -B build -DLIGHTWEIGHT_BUILD_SHARED=OFF   # on Windows, any generator
```

```
CMake Error:
  Error evaluating generator expression:

    $<TARGET_PDB_FILE:Lightweight>

  TARGET_PDB_FILE is allowed only for targets with linker created artifacts.

CMake Generate step failed.  Build files cannot be regenerated correctly.
```

This reproduces even for a consumer that only `add_subdirectory()`/`FetchContent_MakeAvailable()`s Lightweight and never runs `cmake --install` on it — the `install()` call is still *generated*, and generator-expression evaluation happens at generate time regardless of whether install ever runs.

## Suggested fix

Guard the PDB install rule on the actual target type, e.g.:

```cpp
if(WIN32 AND LIGHTWEIGHT_BUILD_SHARED)
    install(FILES $<TARGET_PDB_FILE:Lightweight>
        DESTINATION ${CMAKE_INSTALL_BINDIR}
        CONFIGURATIONS Debug RelWithDebInfo
        COMPONENT DebugSymbols
        OPTIONAL
    )
endif()
```

(or check `get_target_property(_type Lightweight TYPE)` if `LIGHTWEIGHT_BUILD_SHARED` isn't reliably in scope at that point).

## Workaround used downstream

We build Lightweight statically (`LIGHTWEIGHT_BUILD_SHARED=OFF`) to avoid an unrelated dll-interface warning class (C4251/C4275 under `/WX`, since Lightweight's classes carry no dll-export annotations). To make that configuration reach `cmake --build` at all, we currently set `CMAKE_SKIP_INSTALL_RULES=ON` around our `FetchContent_MakeAvailable(Lightweight)` call, which is a broader hammer than we'd like (it skips *all* of Lightweight's install rules, not just this one).

### LINKED ISSUE #545: UPDATE ... RETURNING executes and reports correct row/column counts, but FetchRow() throws 24000 Invalid cursor state (sqliteodbc)
**Problem**

`SqlStatement::Execute()`/`Prepare()` accepts and executes an `UPDATE ... RETURNING ...` statement against sqliteodbc — `SqlResultCursor::NumColumnsAffected()`/`NumRowsAffected()` both report correctly (the driver knows the RETURNING column and the update genuinely applied) — but the first `FetchRow()` throws:

```
24000 (0) - [unixODBC][Driver Manager]Invalid cursor state
```

This blocks any design relying on a single-statement atomic conditional read-modify-return via `RETURNING`.

**Reproducer**

```cpp
Lightweight::SqlStatement stmt;
stmt.ExecuteDirect("CREATE TABLE probe (id INTEGER PRIMARY KEY, n INTEGER NOT NULL)");
stmt.ExecuteDirect("INSERT INTO probe (id, n) VALUES (1, 41)");

stmt.Prepare("UPDATE probe SET n = n + 1 WHERE id = ? RETURNING n");
auto cursor = stmt.Execute(1);
cursor.NumColumnsAffected();  // => 1  (driver knows the RETURNING column)
cursor.NumRowsAffected();     // => 1  (the update genuinely happened)
cursor.FetchRow();            // throws 24000 "Invalid cursor state"
```

Both entry points fail identically (`ExecuteDirect(...)` and `Prepare(...)` + `Execute(...)`), so this is not a prepared-statement binding problem. Controls confirm the failure is specific to `RETURNING`, in the same process, on the same connection:

- a plain `SELECT` prepared/executed the same way fetches normally;
- a plain conditional `UPDATE ... WHERE ...` (no `RETURNING`) reports `NumRowsAffected() == 1`/`== 0` correctly, so the affected-row count is a trustworthy signal on its own.

The driver appears to execute the statement through a non-cursor path and never opens a result set over the returned rows, leaving the statement handle in a state where `SQLFetch` is invalid.

**Environment**

- sqliteodbc via unixODBC 2.3.14
- SQLite 3.53.4
- macOS/arm64
- Lightweight `v0.20260625.0` (vendored)
- SQLite has supported `RETURNING` since 3.35, and the statement itself is valid SQL — the question is whether the driver surfaces the result set.

**Impact**

A consumer needing an atomic conditional read-modify-return (e.g. "decrement a budget counter and return the new row, only if the budget isn't already exhausted") cannot use a single `RETURNING` statement against this driver. The workaround is a two-statement form inside a transaction: a conditional `UPDATE ... WHERE ...` (no `RETURNING`) dispatched on `NumRowsAffected()`, followed by an ordinary read-back by primary key. This is fully equivalent in atomicity (the whole guard clause lives inside the `UPDATE`'s own `WHERE`, evaluated and applied as one indivisible statement under SQLite's write lock) — it just costs one extra round trip and loses the single-statement form's concision.

**What would resolve this**

Either sqliteodbc opening a real cursor/result set for a `RETURNING` statement the way it does for a plain `SELECT`, or a documented note that `RETURNING` result-set fetch is unsupported against this driver so callers know to reach for the two-statement workaround up front instead of discovering it via `FetchRow()`'s exception.

Cross-referenced from `LASTRADA-Software/morph#58`, which tracks the morph-side workaround (`pastebin::PasteModel::execute(const GetPaste&)`, two-statement form).

Yaraslaut:
## Summary

Six independent, self-contained fixes for open issues, each isolated in its own commit for easy review, plus follow-up hardening from review feedback (Copilot + manual C++ review) picked up after the PR was opened. All verified with real `cmake --build`/`ctest` runs, not just read-through.

Closes #535
Closes #536
Closes #542
Closes #543
Closes #544
Closes #545

### #535 — `std::formatter<SqlDynamicBinary<N>>` fails to compile (no `begin()`/`end()`)
The formatter iterated the value with a range-`for` loop, but `SqlDynamicBinary<N>` only exposes `data()`/`size()` — a raw-byte buffer, not an iterable range. Views it via `std::span` instead, keeping the loop range-based. Added regression coverage for printable/non-printable/empty payloads.

A follow-up review pass (this PR) found that `data()`'s const overload originally reached this state via a `const_cast`-and-write on a possibly-const object — undefined behavior per `[dcl.type.cv]`, reproducible with a `const SqlDynamicBinary<8>` local (exactly what the formatter test above uses). Fixed by splitting `data()` into distinct const/non-const overloads: the non-const overload keeps the existing lazy-materialize-buffer behavior, and the const overload returns a static empty-but-non-null sentinel instead of mutating anything, so no `const_cast` and no UB. Added dedicated coverage for both overloads' previously-unexercised branches.

### #536 — `DataMapper::Create` narrows `size_t` to `int` (C2397 under `/WX`)
`CreateInternal` brace-initialized `SqlStatement::LastInsertId()`'s `size_t` return straight into the record's primary key type — a narrowing conversion for any `ServerSideAutoIncrement` PK declared narrower than `size_t` (e.g. `int32_t`, the standard SQL Server `int IDENTITY` mapping). Uses an explicit `static_cast` instead. Added a regression test with an `int32_t` PK — existing coverage only used `uint64_t`, the same width as `size_t`, so the narrowing was never exercised.

### #542 — `SqlStatement.hpp` includes `<Windows.h>` without `NOMINMAX`
`Windows.h`'s `min`/`max` macros leaked into every consumer TU. Guarded all inclusion sites across the library, tools, and test headers with `NOMINMAX`/`WIN32_LEAN_AND_MEAN` (defined only if not already set, so an app-wide override still wins), including sites where an SDK header other than `<Windows.h>` itself (e.g. `<shlobj.h>`) pulls it in transitively. Confirmed `SqlBackup.cpp` and `dbtool/main.cpp` were genuinely broken today via `std::min`/`std::max` calls after their own unguarded include.

### #543 — Windows libzip hard `REQUIRED` with no CPM fallback
Moved libzip resolution to the root `CMakeLists.txt` for every platform, mirroring the `find_package`-then-CPM pattern already used there for `reflection-cpp`/`stdexec`. Verified both paths on Windows: the normal vcpkg path still resolves via `find_package`, and forcing it to fail exercises the CPM fallback and builds/links `Lightweight.dll` against the CPM-built `zip.lib`.

### #544 — Static build fails at generate time (`TARGET_PDB_FILE` invalid)
`$<TARGET_PDB_FILE:Lightweight>` is only valid for targets with a linker-created artifact — a static `.lib` has no PDB, and CMake evaluates generator expressions in `install()` at *generate* time regardless of whether install ever runs. Guarded the rule on `LIGHTWEIGHT_BUILD_SHARED`. Reproduced the exact failure without the fix and confirmed it's gone with the fix, with a clean static `Lightweight.lib` build.

### #545 — `UPDATE ... RETURNING` throws `24000` on sqliteodbc
Investigated rather than blindly patching: this is a real, reproducible platform split, not a flaky driver quirk — unixODBC's sqliteodbc driver (both Ubuntu's official `libsqliteodbc` package and Homebrew's build on macOS) never opens a fetchable cursor over a `RETURNING` result set, while the Windows-native "SQLite3 ODBC Driver" fetches it correctly. `NumRowsAffected()`/`NumColumnsAffected()` are reliable on every driver regardless. Split into two tests: a hard assertion on the row/column-count contract (every platform), and a platform-gated fetch assertion — a hard `REQUIRE` on Windows (verified working), an honest `TryFetchRow()` + `WARN` on the unixODBC-affected platforms rather than papering over the gap with a blanket try/catch. Documented the two-statement portable workaround in `docs/how-to.md`.

## Databases tested
- `sqlite3`, `mssql2017`, `mssql2019`, `mssql2022`, `mssql2025`, `postgres` — full CI matrix green (see checks on this PR)

## Compilers tested
- CI matrix: MSVC `cl` (Debug), `clang-cl` (Debug), Ubuntu Clang 19, Ubuntu GCC 14, macOS (Homebrew LLVM), plus C++26 reflection and C++20 modules (`ddl2cpp`) legs
- Locally: `cl-debug` (MSVC), full suite re-verified against `sqlite3` after each follow-up commit

## Risk assessment
- #535: header/template-only change plus a const-correctness fix in the same type (no `const_cast`, `_base` no longer `mutable`); no ABI impact, additive test coverage.
- #536: header/template-only change, no ABI impact, additive test coverage.
- #542: preprocessor guards only, no behavioral change; defensive on files that don't currently hit the bug.
- #543, #544: build-system only, no source changes; verified both the common (vcpkg) and fallback (CPM) paths.
- #545: no source change to library behavior — test + doc only.

## Code coverage
Codecov flagged patch coverage at 83.33% (2 uncovered lines in `SqlDynamicBinary.hpp`) after the `data()`-overload-split commit. A follow-up commit adds a test exercising both previously-missed branches (the non-const overload's already-materialized-buffer path and the const overload's non-null early return); see the `codecov/patch` check on this PR for the resulting status.


Yaraslaut:
Addressed the four review findings in 4 follow-up commits:

- **`elseif(NOT TARGET zip)` dead condition** (CMakeLists.txt) — replaced with `else()`; no behavior change, re-verified the CPM fallback path still resolves and links.
- **RETURNING doc sample could hit "connection is busy"** (docs/how-to.md) — scoped the cursor so it closes before the read-back statement runs on the same connection.
- **`const_cast` on possibly-const `SqlDynamicBinary`** (SqlDynamicBinary.hpp) — made `_base` `mutable` instead, removing the UB; the formatter's new "empty buffer" test (a genuinely `const` object) now exercises this path without casting away const.
- **RETURNING test couldn't catch its own regression** (SqlStatementDbTests.cpp) — split into two test cases; the fetch assertion is now a hard `REQUIRE` instead of a `WARN`-and-continue, so a regression on the drivers this suite actually runs against (SQLite/PostgreSQL) fails the build instead of passing silently.

Full suite re-verified green on sqlite3: 1348 passed / 1 skipped / 2 failed-as-expected (pre-existing, unrelated clang-cl miscompile placeholders) — same baseline as before, no regressions from any of the fixes.

Yaraslaut:
The last review-fix commit (9d970e65) broke CI: **Sanitizers (asan-ubsan)**, **Sanitizers (tsan)**, **Code coverage**, and **macOS (Homebrew LLVM, SQLite)** all failed at the same line — \`SqlStatementDbTests.cpp:137\`, the hardened \`REQUIRE(cursor.FetchRow())\`.

Root cause confirmed from the CI logs themselves: unixODBC's sqliteodbc driver — both Ubuntu's official \`libsqliteodbc\` apt package and Homebrew's \`sqliteodbc\` on macOS — never opens a fetchable cursor over a \`RETURNING\` result set. \`FetchRow()\` throws \`24000 Invalid cursor state\` there every time, identically to the original issue #545 report (which was also unixODBC/macOS). This isn't a flaky one-off; it hit every non-Windows job that ran the test, while \`Windows-clangcl-debug\` passed — matching what I'd already verified locally on the Windows-native SQLite3 ODBC driver.

Pushed a fix (\`e1a6c6ee\`) gating the hard assertion on \`_WIN32\`: \`REQUIRE\` on Windows (a real regression guard, confirmed working there), tolerant \`TryFetchRow\` + \`WARN\` on every other platform, where the limitation is now CI-confirmed rather than a guess. Waiting on this run to confirm all four jobs go green.

src/tools/shared/Config/ProfileStore.cpp
```diff
@@ -12,6 +12,12 @@
...
 #ifdef _WIN32
     #include <shlobj.h>
+    #ifndef NOMINMAX
+        #define NOMINMAX
+    #endif
+    #ifndef WIN32_LEAN_AND_MEAN
+        #define WIN32_LEAN_AND_MEAN
+    #endif
     #include <windows.h>
 #else
```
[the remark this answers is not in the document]
Yaraslaut: Fixed in ea21f2a6 — moved the NOMINMAX/WIN32_LEAN_AND_MEAN defines ahead of `<shlobj.h>` instead of only in front of the explicit `<windows.h>` below it. Verified `dbtool.exe` still resolves %APPDATA% correctly at runtime after the reorder.

src/Lightweight/DataBinder/SqlDynamicBinary.hpp
```diff
@@ -198,9 +203,13 @@ struct std::formatter<Lightweight::SqlDynamicBinary<N>>: std::formatter<std::str
...
     auto format(Lightweight::SqlDynamicBinary<N> const& text, format_context& ctx) const
     {
+        // SqlDynamicBinary<N> exposes data()/size(), not begin()/end() — it is a raw-byte buffer,
+        // not an iterable range. Walk it by index instead of range-for.
         std::string humanReadableText;
-        for (auto byte: text)
+        auto const* const bytes = text.data();
+        for (std::size_t i = 0; i < text.size(); ++i)
         {
+            auto const byte = bytes[i];
```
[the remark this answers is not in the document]
Yaraslaut: Fixed in ea21f2a6 — replaced the index loop with a range-based `for` over a `std::span` view of `data()`/`size()`, keeping the loop range-based without adding `begin()`/`end()` to `SqlDynamicBinary` itself.

src/Lightweight/DataBinder/Core.hpp
```diff
@@ -3,6 +3,12 @@
 #pragma once
 
 #if defined(_WIN32) || defined(_WIN64)
+    #ifndef NOMINMAX
+        #define NOMINMAX
+    #endif
```
christianparpart, line 8: i think that should be done on cmake level, or what do you think? We generally `NOMINMAX` globally on a project

src/Lightweight/DataBinder/Core.hpp
```diff
@@ -3,6 +3,12 @@
...
+    #ifndef WIN32_LEAN_AND_MEAN
+        #define WIN32_LEAN_AND_MEAN
+    #endif
```
christianparpart, line 11: any chances to move that into the .cpp that need it? I still remember how one listener in our demo once said that we should never included `<Windows.h>` in a public header. Maybe it's worth invesgitating that (sure, we might need to move some inlines into .cpp files for that)

src/Lightweight/DataBinder/Core.hpp
Yaraslaut: I think it is hard to control from cmake how project headers are consumed. So I guess it is hard/impossible to figure out at the time of Lightweight project cmake files loading what will be the flags for the headers

src/Lightweight/DataBinder/Core.hpp
Yaraslaut: last time i tried, i had issues with the sql date types in the headers since they required windows headers, i will create an issue and see what is possible there

Yaraslaut: Opened #547 to track this investigation — scoped it to the specific obstacle (SqlDate/SqlDateTime/SqlTime needing SQL_DATE_STRUCT/SQL_TIME_STRUCT/SQL_TIMESTAMP_STRUCT from the ODBC headers, which need the Windows prelude on this platform) rather than just "move the include," plus the open questions around forward-declaring the struct types and the inlining/perf tradeoff if that code moves into a .cpp. Not blocking this PR.

PR #576, opened 2026-08-19
Add an opt-in prepared-statement cache for repeated SqlStatement executions

### LINKED ISSUE #552: Add prepared-statement caching for repeated SqlStatement executions
## Summary

`SqlStatement::Prepare()` requires the caller to explicitly prepare a statement before execution, and there is no cache that reuses an already-prepared statement handle when the same SQL text is executed again (e.g. across `DataMapper` query-builder calls, or repeated ad-hoc `Execute()` calls with the same query string).

## Motivation

Preparing a statement is a network round-trip to the server for most drivers (MSSQL, PostgreSQL). Frameworks like [userver's uPg driver](https://userver.tech/d7/dc1/md_en_2userver_2pg_2driver.html) transparently convert queries to prepared statements and cache them, so repeat executions of the same query text skip re-preparation. Lightweight currently has no equivalent — every call site that wants this benefit must manage its own `SqlStatement` lifetime and reuse manually.

## Proposal

- Add an opt-in prepared-statement cache, likely owned by `SqlConnection` (or the `Pool`/`DataMapper` layer), keyed by normalized SQL text (+ parameter count/types if needed for correctness across drivers).
- Cache eviction policy: bounded LRU (size configurable), since some drivers/servers cap the number of live prepared statements per connection.
- Must respect per-DBMS behavior via `SqlQueryFormatter`/connection capabilities — e.g. confirm SQLite's ODBC driver prepared-statement semantics don't silently break under reuse (schema changes invalidating a cached plan, etc.).
- Should be transparent to existing call sites: `DataMapper` and the `SqlQuery` DSL should benefit without call-site changes, but a raw `SqlStatement::Prepare()` caller should still be able to opt out.

## Acceptance criteria

- New cache is covered by unit tests in `src/tests/` demonstrating a cache hit avoids re-preparation (e.g. via `SqlLogger` hook counting or a mock/spy).
- Full suite green on sqlite3, mssql2022, postgres.
- Documented in `docs/` (likely `usage.md` or a new `docs/prepared-statement-cache.md`).

## References

- Comparison research against [userver-framework/userver](https://github.com/userver-framework/userver)'s uPg driver, which caches prepared statements automatically.

Yaraslaut:
Closes #552

## What

`SqlConnection` can now keep already-prepared `SQLHSTMT` handles alive in a **bounded LRU pool**, so a
repeat of a query text the connection has seen re-executes a prepared handle instead of preparing a fresh
one.

What that saves is a server-side parse on MS SQL Server and PostgreSQL. Neither driver pays it inside
`SQLPrepare` — measured, `Prepare()` on its own costs ~1.2 µs and sends nothing, even behind a 50 ms link.
The parse rides along with the **first execute** of a freshly prepared handle (`sp_prepexec` on the
Microsoft driver, a Parse/Describe exchange on psqlODBC), and a matching deallocate follows when the
handle is freed. `DataMapper` and the `SqlQuery` DSL build a fresh `SqlStatement` per call site, so the
same handful of query texts paid that pair over and over.

```cpp
auto conn = SqlConnection {};
conn.SetPreparedStatementCacheCapacity(Lightweight::PreparedStatementCacheCapacitySuggested); // 64
// every DataMapper / query-builder / raw Prepare() call on this connection now benefits
```

Pooled applications configure it on the pool instead of per connection:

```cpp
constexpr auto MyPoolConfig = Lightweight::PoolConfig {
    .initialSize = 4,
    .maxSize = 16,
    .growthStrategy = Lightweight::GrowthStrategy::BoundedOverflow,
    .preparedStatementCacheCapacity = Lightweight::PreparedStatementCacheCapacitySuggested,
};
auto pool = Lightweight::Pool<MyPoolConfig> {};
```

## Design

- **`SqlPreparedStatementCache`** (new, owned by `SqlConnection`) — bounded LRU pool keyed by exact SQL
  text. A handle is *checked out* while a statement uses it (`Acquire` removes it, `Release` puts it
  back), so two statements preparing the same text concurrently each get their own handle. Eviction is
  least-recently-released first, which matters because several backends cap live prepared statements per
  session. `std::list` + `unordered_multimap` of `string_view` keys into the stable list nodes, so lookup
  and eviction are both O(1) average.
- **A repeat of the text a statement already holds is served from its own handle**, never routed through
  the pool. Parking the handle only to look that same text straight back up reaches the same handle
  without a prepare, but pays a cursor close, a parameter reset, a hash erase/insert and two string
  allocations for the trip — measured at 10% on MS SQL Server and 6% on SQLite over 1000
  `DataMapper::QuerySingle()` calls. The pool is consulted only when the statement cannot answer from what
  it holds. Those reuses are still counted, as `Statistics::directReuses`, so "no prepare was issued"
  stays observable for every prepare the cache is responsible for.
- **`ExecuteDirect()` / `ExecuteBatchFetch()` park the handle first**, since executing directly discards
  whatever it was prepared for. Without this the `DataMapper`'s `INSERT` handle was thrown away by the
  `SELECT LAST_INSERT_ROWID()` that immediately follows it — the DataMapper test caught exactly that
  (5 misses, 0 hits) before the fix.
- **Releasing pools the count `SQLPrepare` reported** (`m_preparedParameterCount`), not
  `m_expectedParameterCount`, which `BindInputParameter()` overwrites with a "count unknown" sentinel.
  Pooling the sentinel would make every later `Execute(args...)` on that text reject its arguments, and
  re-deriving the count with `SQLNumParams` would put an ODBC call in every release — worth 1.21x → 1.66x
  on `DataMapper::Create()` against MS SQL Server.
- **Configured on the pool, not at every `Acquire()`.** `Pool` creates its mappers with `DataMapper`'s
  default constructor, which goes through `Connect(SqlConnectionString)` and so never reads
  `SqlConnectionDataSource::preparedStatementCacheCapacity` — without a pool-level knob the cache was
  unreachable for pooled code except by repeating a setter at each call site, and mappers created on
  overflow still started cold. `PoolConfig` therefore gains `preparedStatementCacheCapacity` as a
  compile-time policy field alongside the three already there, with
  `LIGHTWEIGHT_POOL_PREPARED_STATEMENT_CACHE_CAPACITY` (default `0`) wiring it to `GlobalDataMapperPool()`.
  The four sites that constructed a pooled `DataMapper` now go through a single `Pool::CreateDataMapper()`
  factory — the place any future pool-wide connection setting belongs too — and the capacity is applied
  under `if constexpr`, so a pool left at the default emits exactly the code it did before the field
  existed.
- **The cache is per connection and cannot be otherwise.** A pooled handle is a child of one connection's
  `SQLHDBC`; it can never be shared with another. Each pooled connection warms up independently, a fully
  warmed pool holds up to `maxSize * capacity` prepared statements on the server, and
  `GrowthStrategy::BoundedOverflow` discards the warmed cache of any connection it destroys on return.
  All three are measured below and called out in the docs, since they drive how the capacity is sized.
- **Opt-in, then transparent.** Capacity defaults to `0` because a pooled handle carries the plan the
  driver derived from the schema at preparation time. Once enabled, every layer on top of `SqlStatement`
  benefits with no call-site change; a single statement opts out via
  `SetPreparedStatementCaching(SqlPreparedStatementCaching::Disabled)`.
- **Invalidation.** `MigrateDirect()` and the `MigrationManager` executor clear the cache after applying a
  script; connect / reconnect / `Close()` clear it as well (statement handles are children of the DBC
  handle). Raw DDL via `ExecuteDirect()` remains the caller's responsibility, documented.
- **Per-DBMS via a connection capability,** not a `switch` in business logic:
  `SqlConnection::SupportsPreparedStatementReuse()` mirrors `SupportsNativeRowBatch()`. On a backend that
  is not `MICROSOFT_SQL` / `POSTGRESQL` / `SQLITE`, a requested capacity is retained but stays inactive,
  so the same setup code is safe to run anywhere.

## Tests

`src/tests/SqlPreparedStatementCacheTests.cpp` — 19 cases, 148 assertions. Cache hits are proven both by
the cache's own hit/miss/eviction/direct-reuse counters and by a `SqlLogger` spy counting logical
`Prepare()` calls: **five prepares of one query text reach the driver once**. Beyond the headline it
covers pooling across statement lifetimes, two concurrent statements each getting their own handle, LRU
eviction at capacity 1, capacity shrink, the per-statement opt-out, the hand-bound parameter-count case,
invalidation by `MigrateDirect()`, `DataMapper::Create()` in a loop (1 miss / 4 hits, no call-site
change), and a differential test that reads the same rows with and without the cache and compares the
results — so a reused handle returning stale or wrong data would fail.

Four cases cover the pool path: a default-configured pool leaves the cache disabled; a configured pool
applies the capacity on **both** non-blocking creation paths and on `BoundedWait`'s own; and a pooled
connection keeps its warmed handles across a return/re-acquire cycle, so five `DataMapper::Create()` calls
split over two acquires still reach the driver exactly once.

### How it was measured

`src/benchmark/prepared_statement_cache.cpp` (target `LightweightPreparedStatementCacheBenchmark`, added
by this PR) runs each workload with the cache off and on and reports the fastest of N repetitions.
`src/benchmark/latency_proxy.py` (also new, stdlib-only, no root) puts the server behind a simulated WAN
link so the same workload can be measured at a realistic round-trip time.

One methodological note worth keeping: **the two settings alternate inside one loop** rather than running
all the "off" repetitions first. A shared database host drifts over the minutes a comparison takes, and a
sequential A-then-B run charged that drift to whichever setting happened to run while the host was busy —
it reported a 20% regression on one workload and a 1.5x win on another that both disappeared once the
settings were interleaved. Prepare counts come from the `SqlLogger` spy and the cache's own counters, so a
scenario that stops exercising the cache shows up as a count, not as a suspiciously small delta.

### One connection (1000 iterations, loopback Docker)

| workload | SQLite 3.49 | PostgreSQL 16.4 | MS SQL Server 2022 |
|---|---|---|---|
| fresh `SqlStatement` per call, prepare + execute + fetch | 1.44x | **4.41x** | **1.68x** |
| fresh `SqlStatement` per call, prepare only | 4.04x | 1.19x | 1.12x |
| one statement, 4 query texts interleaved | 1.29x | **4.15x** | 1.02x |
| `DataMapper::Query<>().Where().All()` | 1.09x | **3.75x** | **1.34x** |
| `DataMapper::Create()` (INSERT + last id) | 0.92x | **1.81x** | **1.30x** |
| `DataMapper::QuerySingle()` by primary key | 1.01x | 1.00x | 1.00x |

Every row confirmed by a second full sweep (within ~5%).

**The saving is absolute, not proportional.** Per query on PostgreSQL a query-builder read goes from
143.5 µs to 38.3 µs — **105 µs removed**. On a 200-row toy table that fixed cost is most of the query,
hence 3.8x; on a query doing real server work the same 105 µs is a smaller share and the ratio falls
toward 1. The number to carry over is the 105 µs, not the multiplier.

`QuerySingle()` gains nothing anywhere: it prepares through the mapper's own long-lived statement, which
reuses its handle either way. On SQLite the whole thing is roughly break-even, and `Create()` is ~8%
slower because its last-insert-id `ExecuteDirect()` has to park the prepared handle — cheaper than a parse
on a network engine, more expensive than one in-process.

### Under a connection pool (500 operations, 3 query shapes per operation)

Worker threads acquire a `DataMapper`, run `QuerySingle()`, a `Query<>().Where().All()` and a `Count()`
through it, and hand it back. Counters are summed across the pool by snapshotting each connection's
`Statistics` at acquire and adding the delta at release — the pool owns the connections and does not
expose them.

| pool shape | SQLite | PostgreSQL | MS SQL Server | `SQLPrepare` cold → warm |
|---|---|---|---|---|
| 1 connection, 1 worker | 1.14x | **2.79x** | **1.35x** | 3 → 0 |
| 1 connection, 4 workers (contended) | 1.17x | **2.71x** | **1.22x** | 3 → 0 |
| 4 connections, 4 workers | 1.13x | **2.53x** | **1.35x** | 12 → 0 |
| 2 idle connections, 8 workers (`BoundedOverflow`) | 1.01x | 1.41x | 1.13x | ~30 → ~24 |

Re-measured after the rebase onto master's pool-health work, so these are taken with `validateOnBorrow`
at its default (`Yes`) — each acquire from the idle set also pays an `IsAlive()` check. Two passes per
database, agreeing within noise.

- The warm-up is exactly **`connections × distinct query texts`** — 3 texts over 4 connections is 12
  prepares, not 3. Nothing is shared between the mappers a pool hands to its workers. But it is paid
  **once**: a pool that keeps its connections converges to zero `SQLPrepare` however many it holds, and
  **pool size does not dilute the steady-state win** (1 and 4 connections reach 2.79x and 2.53x on
  PostgreSQL).
- **Short-lived work does dilute it**: at 6 operations per connection the cold PostgreSQL gain falls from
  2.35x to 1.69x. The cache wants a pool whose connections live across many requests.
- **Anything that retires a connection discards its warmed cache** — `BoundedOverflow` past the idle set,
  but equally `maxIdleTimeMs`, `maxLifetimeMs` and a failed `validateOnBorrow` check, all of which master
  added while this branch was open. The overflow shape never converges: ~24 prepares are still being
  issued after many passes. Not really
  a cache problem: that shape spends most of its time *connecting* (25 ms for 24 operations, versus 2.9 ms
  for the same work on a 4-connection `BoundedWait` pool). Sizing `maxSize` to the real concurrency is
  worth more there than any cache capacity.

### As network latency grows

Loopback is the least favourable setting for the cache: what it removes is round-trips. Time for **one**
`Query<>().Where().All()` behind a simulated link:

| round-trip time | PG off | PG on | speedup | MSSQL off | MSSQL on | speedup |
|---|---|---|---|---|---|---|
| 0 ms (loopback) | 0.14 ms | 0.04 ms | 3.8x | 0.16 ms | 0.12 ms | 1.3x |
| 5 ms | 25.3 ms | 7.6 ms | 3.3x | 12.8 ms | 6.5 ms | 2.0x |
| 10 ms | 41.4 ms | 12.4 ms | 3.3x | 20.9 ms | 10.5 ms | 2.0x |
| 25 ms | 113.5 ms | 34.1 ms | 3.3x | 57.1 ms | 28.6 ms | 2.0x |
| **50 ms** | **201.8 ms** | **60.5 ms** | **3.3x** | **101.8 ms** | **51.0 ms** | **2.0x** |

Every scenario at 50 ms, per call:

| scenario | PG off | PG on | MSSQL off | MSSQL on |
|---|---|---|---|---|
| fresh statement, prepare + execute + fetch | 201.7 ms | 60.5 ms | 101.2 ms | 50.6 ms |
| one statement, 4 interleaved texts | 201.7 ms | 90.8 ms | 55.9 ms | 50.7 ms |
| `Query<>().Where().All()` | 201.8 ms | 60.5 ms | 101.8 ms | 51.0 ms |
| `QuerySingle()` | 50.6 ms | 50.5 ms | 50.8 ms | 50.8 ms |
| `Create()` | 307.7 ms | 166.5 ms | 167.4 ms | 116.7 ms |
| `Prepare()` with no execute | 0.002 ms | 0.002 ms | 0.002 ms | 0.002 ms |

The speed-up is **flat across the whole range**, so the cost is purely round-trips. Dividing by the
round-trip time counts them:

| | PostgreSQL | MS SQL Server |
|---|---|---|
| query-builder read, no cache | **≈4.0 round-trips** | **≈2.0 round-trips** |
| query-builder read, cached | ≈1.2 | ≈1.0 |
| **saved per query** | **≈2.8 round-trips** | **exactly 1 round-trip** |
| `QuerySingle()`, either way | 1.01 | 1.02 |
| `Prepare()` with no execute | **0** | **0** |

- **One round-trip is the floor, and `QuerySingle()` is already on it** — 1.01 round-trips behind a 50 ms
  link with the cache on *or* off. That is why it never gains anything, on any backend.
- **`Prepare()` costs no round-trip at any latency.** This disproves the claim this PR originally made —
  that `SQLPrepare` itself was the round-trip — and the docs *and* the `SqlPreparedStatementCache` /
  `SqlConnectInfo` header comments were corrected accordingly.
- **The two backends save different things.** psqlODBC spends ~3 extra round-trips per freshly prepared
  statement (Parse/Describe plus the deallocate when the handle goes away); the Microsoft driver folds the
  prepare into `sp_prepexec` and spends exactly one — the `sp_unprepare` after a short-lived handle. That
  also explains why reusing *one* statement across four query texts gains nothing on MS SQL Server
  (1.10x): no handle is torn down, so there is no round-trip to save.

**Bottom line:** on loopback this is a micro-optimisation; across an availability zone or a region it
removes most of the query — at 50 ms a PostgreSQL query-builder read goes from 202 ms to 61 ms. Off by
default, so **no change to existing behaviour or performance** unless a capacity is set.

### What the measurements changed in the implementation

Two defects were found by benchmarking and are fixed in the branch (both folded into the first commit, so
the history never ships the slower version):

- **`SQLNumParams` on every release** — the parameter count was re-derived from the driver on each release
  though it was already known. `DataMapper::Create()` on MS SQL Server: 1.21x → 1.66x.
- **A same-text re-prepare routed through the pool** — same handle, no prepare, but a cursor close, a
  parameter reset and two string allocations for the trip. It made `QuerySingle()` **10% slower on MS SQL
  Server and 6% slower on SQLite with the cache enabled** than with it off; now 1.00x.

## Risk assessment

**Low–medium**, mitigated by the feature being off by default: with capacity `0` the new code paths are
inert and the diff is behaviour-neutral.

- *Stale plans after DDL* — the real hazard. Mitigated by clearing on migrations and on connect/close, and
  documented for raw `ExecuteDirect()` DDL.
- *Handle-state leakage between uses* — a handle is returned neutral: `CloseCursor()` (which also tears
  down block-prefetch still referencing it), `SQL_UNBIND`, `SQL_RESET_PARAMS`, and the parameter-array
  attribute reset. None of these unprepare the statement.
- *Server-side handle count under a pool* — the bound is per connection, so a pool multiplies it by
  `maxSize`. Default `0` keeps this inert; the sizing guidance is in `docs/best-practices.md`.
- *Per-DBMS* — gated by `SupportsPreparedStatementReuse()`; unverified backends never pool.
- *Threading* — the cache is per-connection and not thread-safe, matching `SqlConnection` itself; the
  async layer already serializes ODBC work per connection on a strand. The pooled benchmark drives eight
  worker threads through it.
- *ABI* — new member on `SqlStatement`, new fields in `SqlConnection::Data` (opaque, heap-allocated), and
  a new member on `PoolConfig`, which changes the NTTP mangling of every `Pool<Config>` instantiation.
  Source-compatible throughout (the new `PoolConfig` field is defaulted, so existing designated-initializer
  call sites are untouched); not ABI-compatible, consistent with how this repo evolves.

## Databases tested

Full suite, all green, run under **both** `clang-debug` and `gcc-release`:

| env | cases | assertions | skipped |
|---|---|---|---|
| **sqlite3 (3.49.1)** | 1540 | 14607 | 1 (pre-existing `AlterTable AlterColumn` SQLite skip) |
| **mssql2022 (Docker)** | 1540 | 14582 | 3 (pre-existing per-DBMS skips) |
| **postgres (Docker 16.4)** | 1540 | 14541 | 2 (pre-existing per-DBMS skips) |

## Compilers tested

- **clang-debug** (PEDANTIC + ASan/UBSan) — clean build, all three databases green. `clang-tidy` could not
  be run to completion on this host: the locally installed clang-tidy is newer than CI's and flags
  `readability-redundant-typename` / `readability-function-cognitive-complexity` in `DataMapper/Record.hpp`
  and `DataMapper/DataMapper.hpp`, files this branch does not touch. Pre-existing, unrelated to this PR —
  but it does mean this branch's own clang-tidy cleanliness rests on CI.
- **gcc-release** (GCC 15, `-O2`, Linux) — clean build, and the full suite re-run against all three
  databases with it. That is the binary CI's database legs actually execute (`AGENT.md`).
- **C++20 modules / C++26 reflection** — not run. This change adds no namespace-scope entity to a public
  header (the new counter is a data member, the new `RecordDirectReuse()` an inline member function), so
  the module-linkage rule in `AGENT.md` does not apply to it.

## Docs

`docs/usage.md` gains a "Prepared-statement cache" section next to the block-prefetch one it mirrors —
the pool configuration, the CMake option, and a **"What it is worth, measured"** section carrying all
three tables above and the commands that reproduce them. `docs/best-practices.md` gains a sizing/opt-out
entry, the headline numbers, the `maxSize * capacity` budgeting note, and the guidance that the cache is
for network-backed engines. `scripts/check-doc-snippets.py` passes (41 snippets).

## Commit layout

```
SqlStatement: add an opt-in prepared-statement cache on SqlConnection
Pool: configure the prepared-statement cache from PoolConfig
bench: measure the cache against the same workload without it
bench: measure the cache under a connection pool
bench: measure the cache across simulated network latency
docs: adapt the cache's documentation to the pool's health controls
```

Rebased onto master (through #575). Three conflicts, none mechanical:

- **`~SqlStatement`** — master added an early return when the connection is already closed. That guard now
  precedes the release-to-cache, deliberately: a handle whose connection is gone must not be handed to
  that connection's cache either.
- **`PoolConfig` / creation sites** — master introduced its own `MakeEntry()` factory, which supersedes the
  `CreateDataMapper()` this branch had added for the same purpose. `CreateDataMapper()` is gone and the
  capacity is applied inside `MakeEntry()`; `preparedStatementCacheCapacity` joins master's
  `validateOnBorrow` / `maxIdleTimeMs` / `maxLifetimeMs` in `PoolConfig`.
- **Docs** — `docs/connection-pool.md` is new on master and is now the canonical `PoolConfig` page, so the
  field is documented there too, including how retirement interacts with a warmed cache.

The branch was rewritten from 19 commits into these, then rebased. Both commits that touch library code were built
from scratch individually and the cache suite run at the second, so the branch is bisectable. Verified by
diffing the pre-rewrite backup against the new head: identical **except** three `using`-declarations in
`Lightweight.cppm` that the repo's pre-commit `clang-format` alphabetised. Those three come from master's
`SqlStatistics` work and were already out of order there; the hook re-sorted them when the file was
restaged. Formatting only, no semantic change — say the word if you would rather it went in separately.

https://claude.ai/code/session_01SioJ1puwMnaAz7g1TjEGtR


christianparpart (approved):

christianparpart:
@Yaraslaut do we actually have a meassured performance improvement? Can you try that, please, and provide the numbers? It's possible that it is not worth doing this. 🤔

Yaraslaut:
Good question — it did not, and measuring it changed the PR. There is now a benchmark
(`LightweightPreparedStatementCacheBenchmark`, `src/benchmark/prepared_statement_cache.cpp`) and real
numbers, plus two fixes it uncovered.

### The numbers

1000 iterations, `clang-release`, Docker-local servers. Each workload runs with the cache off and on,
**alternating the two settings inside one loop** — a sequential A-then-B run on this host was worthless,
it reported a 20% regression and a 1.5x win that both vanished once the drift was shared. Fastest of 11
repetitions. Speed-up = `off_ms / on_ms`, so >1 is faster with the cache.

| workload | SQLite 3.49 | PostgreSQL 16.4 | MS SQL Server 2022 |
|---|---|---|---|
| fresh `SqlStatement` per call, prepare + execute + fetch | 1.44x | **4.41x** | **1.68x** |
| fresh `SqlStatement` per call, prepare only | 4.04x | 1.19x | 1.12x |
| one statement, 4 query texts interleaved | 1.29x | **4.15x** | 1.02x |
| `DataMapper::Query<>().Where().All()` | 1.09x | **3.75x** | **1.34x** |
| `DataMapper::Create()` (INSERT + last id) | 0.92x | **1.81x** | **1.30x** |
| `DataMapper::QuerySingle()` by primary key | 1.01x | 1.00x | 1.00x |

Second sweep for repeatability, same order: SQLite 1.42 / 4.10 / 1.32 / 1.08 / 0.93 / 1.00, PostgreSQL
4.27 / 1.20 / 3.89 / 3.73 / 1.96 / 1.00, MS SQL Server 1.70 / 1.12 / 0.99 / 1.35 / 1.32 / 0.98.

### Is it worth it?

**For PostgreSQL and MS SQL Server, yes.** The row that matters is
`DataMapper::Query<>().Where().All()` — the query-builder read path applications actually run: **3.75x**
on PostgreSQL, **1.34x** on MS SQL Server. psqlODBC prepares server-side, so a re-prepare is a genuine
round-trip; the Microsoft driver folds prepare into `sp_prepexec`, so its win comes from keeping one
handle alive (`sp_execute`) rather than preparing and unpreparing a fresh one per call site.

**For SQLite, roughly break-even** — no network, no round-trip to save, and `DataMapper::Create()` is ~8%
slower because its last-insert-id `ExecuteDirect()` has to park the prepared handle. That is now written
down in the docs as "enable it for the network-backed engines".

`QuerySingle()` gains nothing anywhere: it prepares through the mapper's own long-lived `_stmt`, which
already reuses its handle for a repeat of the same text with or without the cache.

### Two things the measurements found, both fixed

Each is a separate commit, so either can be dropped.

- **fe055aca — `SQLNumParams` on every release.** Handing a handle back to the pool re-asked the driver
  for its parameter count whenever the caller had hand-bound parameters, though the count was already in
  `m_preparedParameterCount`. Removing it took `DataMapper::Create()` on MS SQL Server from 1.21x to
  1.66x.
- **4dec5a89 — a same-text re-prepare went through the pool.** Since 9c8b8223, asking a statement for the
  query text it already held parked the handle and looked that same text straight back up: same handle,
  no `SQLPrepare`, but a cursor close, a parameter reset and two string allocations for the trip. That
  made `QuerySingle()` **10% slower on MS SQL Server and 6% slower on SQLite with the cache enabled than
  with it off** — the feature actively hurting its most common caller. It is served from the held handle
  again, and what that routing was protecting (the reuse being visible rather than the counters reading
  0) is kept as a new `Statistics::directReuses`.

### Verification

Full suite green on sqlite3 / mssql2022 / postgres under **both** `clang-debug` and `gcc-release`
(1540 cases; 14607 / 14582 / 14541 assertions). PR description updated with the table, the docs carry it
too plus the command to reproduce.

Yaraslaut:
Measured. The benchmark now has a pooled section: worker threads acquire a `DataMapper` from a `Pool`,
run three distinct query shapes through it (`QuerySingle()`, `Query<>().Where().All()`, `Count()`) and
hand it back. Cache counters are summed across the pool by snapshotting each connection's `Statistics`
at acquire and adding the delta at release, and reported for the **cold** pass and a **warm** one
separately — the warm-up is the whole question here.

500 operations (1500 queries), 7 repetitions, same interleaved off/on method:

| pool shape | SQLite | PostgreSQL | MS SQL Server | `SQLPrepare` cold → warm |
|---|---|---|---|---|
| 1 connection, 1 worker | 1.13x | **2.30x** | **1.21x** | 3 → 0 |
| 1 connection, 4 workers (contended) | 1.13x | **2.18x** | **1.21x** | 3 → 0 |
| 4 connections, 4 workers | 1.13x | **2.18x** | **1.40x** | 12 → 0 |
| 2 idle connections, 8 workers (`BoundedOverflow`) | 1.02x | 1.48x | 1.11x | 30 → 24 |

### Your intuition is right, and the counters show it exactly

**The warm-up is `connections × distinct query texts`.** Three query texts over four connections is
**12** `SQLPrepare` calls, not 3 — the counters land on that number precisely. Nothing is shared, just as
you said.

But that is a **one-off**, not a per-query tax: the warm column is **0** for every pool that keeps its
connections, whether it holds one or four. **Pool size does not dilute the steady-state win** — 1 conn
and 4 conns both reach 2.18x on PostgreSQL. More connections buy more warm-up, not more per-query cost.

**Where it does bite is short-lived work.** Re-run with only 24 operations (6 per connection), the cold
pass on PostgreSQL:

| pool shape | 500 ops | 24 ops |
|---|---|---|
| 4 connections, 4 workers | 2.09x | **1.40x** |

The 12 prepares are still being paid off at that point. So the cache wants a pool whose connections live
across many requests — which is the normal case, but worth stating.

### The real trap is `BoundedOverflow`, not the worker count

A connection created past the idle set is **destroyed when returned**, and its warmed cache with it. That
pool never converges: **24 `SQLPrepare` calls are still being issued after many passes**, and the gain
falls from 2.18x to 1.48x on PostgreSQL and to nothing on SQLite.

Worth noting it is not the cache that hurts there — that shape spends most of its time *connecting*
(8 workers, 2 idle slots: 25 ms for 24 operations versus 2.9 ms for the same work on a 4-connection
`BoundedWait` pool). **If the pool overflows, raising `maxSize` to cover the real concurrency is worth far
more than any cache capacity.**

### Why the pooled numbers are lower than the single-connection ones

2.2x here versus 3.75x for the bare `Query<>().Where().All()`: the mix includes `QuerySingle()`, which
gains nothing anywhere (it reuses the mapper's own `_stmt`). A blended workload is the honest number.

Both docs pages carry all of this, plus the command to reproduce it. Pushed as 5518f713.

Yaraslaut:
Two answers: what the ratios actually are, and what happens as latency grows.

## What "2x" means

It is **wall-clock for the whole workload**, so 2.18x = it finished in 46% of the time. But the cache
does not make a query proportionally faster — it removes a **fixed** per-query overhead. Per query, on
loopback (1000 iterations, single connection):

| scenario | PostgreSQL off → on | saved | MS SQL off → on | saved |
|---|---|---|---|---|
| `Query<>().Where().All()` | 143.5 → 38.3 µs | **105 µs** | 163.5 → 121.6 µs | **42 µs** |
| fresh statement, prepare+execute+fetch | 140.6 → 31.9 µs | 109 µs | 91.3 → 54.5 µs | 37 µs |
| `Prepare()` alone, no execute | 1.31 → 1.11 µs | ~0 | 1.19 → 1.06 µs | ~0 |

So on a 200-row toy table that fixed ~105 µs is most of the query, hence 3.8x. On a query that does real
server work the same 105 µs is a smaller share and the ratio falls toward 1. **The saving is absolute,
not proportional** — that is the number to carry over, not the multiplier.

Note the third row: `Prepare()` on its own costs ~1.2 µs and sends **nothing**. Neither driver parses at
`SQLPrepare`; the parse rides along with the first execute. I corrected the docs, which claimed otherwise.

## Under latency

`src/benchmark/latency_proxy.py` (new, stdlib-only, no root) puts the server behind a simulated WAN link.
Swept 0 → 50 ms round-trip. Time for **one** `Query<>().Where().All()`:

| RTT | PG off | PG on | speedup | MSSQL off | MSSQL on | speedup |
|---|---|---|---|---|---|---|
| 0 ms (loopback) | 0.14 ms | 0.04 ms | 3.8x | 0.16 ms | 0.12 ms | 1.3x |
| 5 ms | 25.3 ms | 7.6 ms | 3.3x | 12.8 ms | 6.5 ms | 2.0x |
| 10 ms | 41.4 ms | 12.4 ms | 3.3x | 20.9 ms | 10.5 ms | 2.0x |
| 25 ms | 113.5 ms | 34.1 ms | 3.3x | 57.1 ms | 28.6 ms | 2.0x |
| **50 ms** | **201.8 ms** | **60.5 ms** | **3.3x** | **101.8 ms** | **51.0 ms** | **2.0x** |

Full per-scenario data:

| scenario (per call) | RTT | PG off | PG on | MSSQL off | MSSQL on |
|---|---|---|---|---|---|
| fresh stmt, prep+exec+fetch | 50 ms | 201.7 ms | 60.5 ms | 101.2 ms | 50.6 ms |
| 1 stmt, 4 interleaved texts | 50 ms | 201.7 ms | 90.8 ms | 55.9 ms | 50.7 ms |
| `Query<>().Where().All()` | 50 ms | 201.8 ms | 60.5 ms | 101.8 ms | 51.0 ms |
| `QuerySingle()` | 50 ms | 50.6 ms | 50.5 ms | 50.8 ms | 50.8 ms |
| `Create()` | 50 ms | 307.7 ms | 166.5 ms | 167.4 ms | 116.7 ms |
| `Prepare()`, no execute | 50 ms | 0.002 ms | 0.002 ms | 0.002 ms | 0.002 ms |

**The speed-up is flat across the entire range**, which means the cost is purely round-trips. Divide by
the RTT and the table becomes a count of them:

| | PostgreSQL | MS SQL Server |
|---|---|---|
| query-builder read, no cache | **≈4.0 round-trips** | **≈2.0 round-trips** |
| query-builder read, cached | ≈1.2 | ≈1.0 |
| **saved per query** | **≈2.8 round-trips** | **exactly 1 round-trip** |
| `QuerySingle()`, either way | 1.01 | 1.02 |
| `Prepare()` with no execute | **0** | **0** |

### What this explains

- **One round-trip is the floor, and `QuerySingle()` is already on it** — 1.01 round-trips at 50 ms with
  the cache on *or* off. That is the whole reason it never gains anything, on any backend.
- **`Prepare()` never talks to the server** — 1.2 µs even behind a 50 ms link.
- **The two backends save different things.** psqlODBC spends ~3 extra round-trips per freshly prepared
  statement (Parse/Describe + the deallocate when the handle goes away). The Microsoft driver folds the
  prepare into `sp_prepexec` and spends exactly **one** extra — the `sp_unprepare` after a short-lived
  handle. That also explains the one result that looked odd earlier: reusing *one* statement across four
  query texts gains nothing on MS SQL Server (1.10x) because no handle is ever torn down, so there is no
  round-trip to save.

### Bottom line

Across an availability zone (~1–2 ms) or a region (~25–50 ms) the cache is worth far more than the
loopback numbers suggest: **at 50 ms a PostgreSQL query-builder read goes from 202 ms to 61 ms.** On
loopback it is a micro-optimisation; over a WAN it removes most of the query.

Pushed as 80cabc41 (proxy + sweep), 3c4cc6ae (docs correction). Both docs pages carry the tables and the
reproduction commands.

Yaraslaut:
**Branch rewritten** (force-pushed) — 19 commits regrouped into 5 atomic ones, and the PR description now
carries all three measurement tables.

```
352e9929  SqlStatement: add an opt-in prepared-statement cache on SqlConnection
340d559f  Pool: configure the prepared-statement cache from PoolConfig
950f594c  bench: measure the cache against the same workload without it
b55b8a91  bench: measure the cache under a connection pool
98c78d6d  bench: measure the cache across simulated network latency
```

One thing worth flagging: **the two performance fixes I posted earlier no longer exist as separate
commits.** Both fixed code that only ever lived on this branch, so rewriting from the net diff folds them
into commit 1 — a reader of the history never sees the slower version. I said earlier that each could be
dropped independently; that is no longer true. The reasoning behind both decisions survives in the code
comments and in commit 1's message:

- releasing pools `m_preparedParameterCount` rather than re-deriving the count with `SQLNumParams`;
- a repeat of the text a statement already holds is served from its own handle and never routed through
  the pool, counted as `Statistics::directReuses`.

Verification of the rewrite: `git diff` between the pre-rewrite backup and the new HEAD is **empty except
for three `using`-declarations in `Lightweight.cppm`** that the repo's pre-commit clang-format
alphabetised. Those three lines come from master's `SqlStatistics` work and were already out of order
there; the hook re-sorted them when the file was restaged. Formatting only, no semantic change — but it
does mean this branch now carries that one-hunk tidy-up of master's code, which I did not intend and can
strip if you would rather it went in separately.

Also corrected in the rewrite: the `SqlPreparedStatementCache` and `SqlConnectInfo` header comments still
claimed `SQLPrepare` costs a round-trip. The latency sweep measures it at ~1.2 µs behind a 50 ms link, so
they now say what actually happens — the parse rides with the first execute.

Yaraslaut:
Correction to my previous note — the first rewrite had a bug and has been re-pushed. The commit SHAs
below are the current ones.

```
bca6f164  SqlStatement: add an opt-in prepared-statement cache on SqlConnection
7e7abcc0  Pool: configure the prepared-statement cache from PoolConfig
327991a5  bench: measure the cache against the same workload without it
7ea014d7  bench: measure the cache under a connection pool
7c3efc00  bench: measure the cache across simulated network latency
```

**What was wrong:** splitting `SqlPreparedStatementCacheTests.cpp` between commits 1 and 2, I moved the
contiguous block of pool tests but missed one that sits at the end of the file, after the non-pool cases.
Commit 1 therefore kept a test using `CachingWaitPoolConfig` while the config itself had moved to commit
2 — commit 1 did not compile, so the branch was not bisectable. Caught by building each commit rather
than trusting the split; fixed by moving that test to commit 2 as well.

Verified now:

- `bca6f164` and `7e7abcc0` each configure and build clean from scratch (Clang, Release, tests on), and
  the cache suite passes at `7e7abcc0` (19 cases, 148 assertions).
- `git diff` from the pre-rewrite backup to the new HEAD is empty apart from the three alphabetised
  `using`-declarations in `Lightweight.cppm` described in my previous comment.
- Full suite green again on HEAD: sqlite3 1540/14607, mssql2022 1540/14582, postgres 1540/14541.

PR #581, opened 2026-08-19
Add an opt-in SqlStatistics collector with a Tracy sink

### LINKED ISSUE #555: Add a built-in statistics/metrics collector on top of SqlLogger hooks
## Summary

`SqlLogger` exposes per-event hooks (`OnExecute`, `OnExecuteBatch`, `OnFetchRow`, `OnFetchBlock`, `OnConnectionIdle`, `OnConnectionReuse`, etc.), but there is no built-in aggregation on top — no counters, latency histograms, or a structured statistics snapshot. Every caller who wants metrics (query counts, p50/p99 latency, pool utilization, connection churn) has to implement their own aggregator against the logger hooks from scratch.

## Motivation

userver exposes structured, ready-made statistics for its DB drivers (e.g. `storages::postgres::ClusterStatistics` — per-host connection counts, latencies, master/replica/sync-replica breakdowns) that integrate with its metrics/exporter system out of the box. Lightweight's `SqlLogger` hooks (landed via #548) are the right seam to build this on, but today they're a pure callback interface with no default consumer.

## Proposal

- Add an optional, built-in statistics collector — e.g. `SqlStatistics` or `SqlMetricsCollector` — that implements `SqlLogger` and aggregates:
  - execution counts (by statement / by outcome: success, error, retried),
  - latency distribution (min/max/avg, or histogram buckets) for `Execute`/`ExecuteBatch`/fetch operations,
  - pool-level counters building on the existing `OnConnectionIdle`/`OnConnectionReuse` hooks (active/idle connection counts, acquire wait time, reuse rate).
- Expose a snapshot/read API (`SqlStatistics::Snapshot()` or similar) returning a plain struct — data-driven, no framework-specific exporter baked in, so users can wire it into Prometheus/StatsD/whatever they use.
- Should be zero-cost when not enabled (opt-in, since `SqlLogger` hooks are already zero-cost when no logger is installed).
- Consider whether this belongs at the `SqlConnection` level, the `Pool` level, or both (pool-level stats need connection-level stats aggregated across the pool).

## Acceptance criteria

- New collector type covered by unit tests: verify counts/latencies are captured correctly across `Execute`, `ExecuteBatch`, and pool acquire/release cycles.
- No measurable overhead when the collector isn't installed (existing `SqlLogger` no-op path preserved).
- Full suite green on sqlite3, mssql2022, postgres.
- Documented in `docs/` and cross-referenced from wherever `SqlLogger` is currently documented.

## References

- Comparison research against [userver-framework/userver](https://github.com/userver-framework/userver)'s `storages::postgres::ClusterStatistics`.
- Builds on existing hook points from #548 (`SqlLogger::OnConnectionIdle`/`OnConnectionReuse` wiring).

Yaraslaut:
Closes #555.

## What and why

`SqlLogger` exposes per-event hooks but nothing aggregates them, so every caller wanting query counts, latency percentiles or pool utilization has to build an aggregator from scratch.

While implementing this I found that **three of the four things the issue asks for are not observable through the `SqlLogger` hooks at all**:

- `OnExecute` fires *before* the ODBC call and there is no matching "finished" hook, so latency cannot be measured (`StandardLogger` approximates it by treating `OnFetchEnd` as the stop marker, which conflates execute+fetch and only fires when a result set is consumed).
- `OnConnectionReuse` fires *after* a mapper is already in hand, so pool acquire wait time is invisible — neither the blocking `cv.wait` in `Acquire()` nor the parked-coroutine path emits anything.
- The `BoundedOverflow` return path destroys an over-capacity mapper with no event at all, so an idle count reconstructed from hooks drifts.

Rather than growing the public virtual interface every deriver implements, collection is instrumented at the implementation sites and **gated at build time**, so a release build carries none of it — not a no-op virtual call, but no code.

```sh
cmake -B build -D LIGHTWEIGHT_ENABLE_STATISTICS=ON
```

## API

`SqlStatistics::Snapshot()` returns a plain struct — no exporter concepts baked in:

- per-operation counters (`succeeded` / `failed` / `retried`) for Execute, ExecuteDirect, ExecuteBatch, Prepare, PoolAcquire
- 32-bucket power-of-two latency histograms with exact `AverageMicroseconds()` and `PercentileMicroseconds(p)`
- pool acquire/release/reuse/discard counts, live occupancy, and a wait-time distribution sampled only from acquisitions that actually blocked
- connection churn, and `rowsFetched` vs `blockFetches` (N rows collapsing into ceil(N/depth) round-trips)

Recording is lock-free relaxed atomics. Failure is inferred from `std::uncaught_exceptions()` in the scope destructor rather than a flag, so every path that throws out of `RequireSuccess` is classified without its call site saying so.

The macros follow the `TracyProfiler.hpp` convention exactly: with the option off they consume their arguments and expand to nothing. `IsEnabled()` is `constexpr` and the types exist in both configurations, so downstream metrics code compiles either way without `#ifdef`.

## Tracy

With `LIGHTWEIGHT_ENABLE_TRACY=ON` as well, every sample is additionally emitted as a Tracy plot (`Sql.Execute.us`, `Sql.Pool.Idle`, …), so one set of instrumentation feeds both the snapshot API and the Tracy GUI.

A new **`tracy_capture` CI job** exercises that combination end to end: it builds Tracy's headless `tracy-capture`/`tracy-csvexport` (pinned to v0.13.1, matching the CPM fallback), runs the Chinook example with both switches on against SQLite, then asserts **independently** that (a) the collector reported non-zero counters with real latency samples and (b) the `.tracy` file contains the expected Lightweight zones. Trace, CSV and log upload as artifacts.

## Documented limitations

Called out in `docs/statistics.md` and the headers rather than papered over:

- `pool.checkedOut` is only maintained by `GrowthStrategy::BoundedWait`; the non-blocking strategies never track one, so it reads 0 there.
- `retried` has no producer yet — master has no transparent-retry path. The API and tests exist; it stays at zero until a retry policy lands.
- A snapshot is not atomic as a whole (relaxed atomics); counters can disagree by a few samples at the edges.

## Performance impact

Zero when the option is off — nothing is compiled. When on, an instrumented operation costs one `steady_clock::now()` pair plus a few relaxed atomic increments, negligible against an ODBC round-trip. The full suite showed no measurable runtime difference between the enabled and disabled builds.

## Risk assessment

- **Per-DBMS**: none — no SQL text, no type binding, no formatter changes. Verified identical pass counts across all three databases.
- **Threading**: recording is lock-free and safe from any thread; a dedicated 8-thread × 1000-sample contention test covers the min/max CAS loops. Pool occupancy is only read under `_mutex`; the async `await_resume` path deliberately records wait latency but *not* occupancy, since it runs without the lock.
- **ABI**: new public header and one new module export (`ToStringView`); no existing signature changed.
- **`SqlLogger`**: untouched — installing statistics does not displace a user's logger.

## Databases tested

- `sqlite3` — 1416 passed, 1 skipped, 13716 assertions
- `postgres` (Docker 16.4) — 1415 passed, 2 skipped, 13641 assertions
- `mssql2022` (Docker, SQL Server 16.00.4265) — 1414 passed, 3 skipped, 13688 assertions
- `sqlite3` with `LIGHTWEIGHT_ENABLE_STATISTICS=OFF` — 1416 passed, 1 skipped, 13677 assertions

## Compilers tested

- Apple Clang 17 debug, statistics ON and OFF — full suite green in both
- Homebrew GCC 15.3 — compiled the new and instrumented translation units; a **full** GCC build is not possible on macOS because `SqlLogger.cpp` (untouched by this PR) uses `std::stacktrace`, which Homebrew GCC does not provide. CI's GCC 14 / `gcc-release` leg is the real check.
- `LIGHTWEIGHT_BUILD_MODULES=ON` and the C++26 reflection leg are not reproducible locally; relying on CI. The module export of `ToStringView` (needed because the Chinook example now calls it, and `ddl2cpp` builds Chinook with modules on) is the change most likely to need a follow-up.

## Coverage

New `src/tests/SqlStatisticsTests.cpp`: 15 cases / 87 assertions covering bucket mapping and saturation, empty-histogram behaviour, percentile estimation, per-operation counts and latency, reset, pool acquire/release cycles, rows-vs-block-fetch separation, exception-based failure classification, multi-threaded recording, and end-to-end capture through real Execute / ExecuteDirect / ExecuteBatch / pool paths. The same 15 cases run in a statistics-disabled build and assert the zero-read path.


christianparpart (approved):
i love the idea 👍

PR #591, opened 2026-08-21
SqlMigration: diff registered migrations against the live schema, and generate migrations from it

### LINKED ISSUE #564: Migrations: no autogenerate from model diff (both halves already exist in-tree)
The migration engine is ahead of Alembic, EF Core and Rails on several axes — checksums with drift detection, a dependency DAG with topological ordering, release tagging, structured per-step failures, plugin distribution. The one capability all three of those have and Lightweight does not is **autogenerate**: diff the declared model against the current schema and emit the migration.

### Why this is unusually cheap here

Both halves already ship:

- **Live schema → model**: `SqlSchema` + `CxxModelPrinter` already read a live database and produce record definitions, including relation inference (`src/Lightweight/Tools/CxxModelPrinter.cpp`, `docs/ddl2cpp-relation-generation.md`).
- **Registered migrations → end state**: `FoldRegisteredMigrations` folds every registered `Up()` into a per-table end state with `creationOrder`, `indexes` and chronological `dataSteps` (`SqlMigration.hpp:536`, doc block `:458-535`). It is live code, already used by `hard-reset` and `unicode-upgrade-tables` (`SqlMigration.cpp:2700, 3008`).

What is missing is the comparison between the two and an emitter for a timestamped `LIGHTWEIGHT_SQL_MIGRATION` block.

### Suggested scope

1. `dbtool diff` — compare `FoldRegisteredMigrations(...)` output against the live schema and print the delta as a migration plan.
2. `dbtool generate <title>` — the same, emitted as a `.cpp` migration source file with a `YYYYMMDDHHMMSS` timestamp. This is Alembic's `revision --autogenerate` / EF's `Add-Migration`.
3. Scaffolding-only mode (empty migration with the timestamp filled in) as the trivial first step, since hand-writing the timestamp is the current friction point.

### Related loose end

`SqlMigration.hpp:465` documents a `dbtool fold` command that emits a self-contained baseline (`.cpp` plugin or `.sql`). No such command exists in `src/tools/dbtool/main.cpp`. Either implement it or drop the reference — squash-to-baseline is the natural companion to autogenerate.

### Second, smaller gap

`MigrationBase::ComputeChecksum` hashes the **rendered SQL** (`SqlMigration.cpp:2149-2165`), so the same migration checksums differently against SQLite and SQL Server — which is why `rewrite-checksums` has to exist. Hashing the migration *plan* structure instead would make checksums portable across dialects and remove a class of false drift.

Yaraslaut:
Closes #564

> Opened as a **draft**: the feature works end to end and the suite is green, but it is the author's work-in-progress and has not had a design pass. See *Status* below.

## What & why

Adds the schema-drift half of #564 — what the registered migrations *intend* versus what the database actually *has*.

`DiffAgainstLiveSchema` folds the registered migrations into the schema they describe, reads the live catalog, and reports the delta as owned, freely copyable values:

- tables the migrations declare that the database lacks, as ready-to-replay `CREATE TABLE` plans **in creation order**, so foreign keys resolve when the plan is replayed front to back;
- tables the database has that no registered migration declares (reported only — autogenerate never emits drops);
- per-column drift in rendered type or nullability.

`MakeSchemaDiffPlan` turns the same delta into replayable plan elements on demand rather than storing them, because the plan element types hold `string_view`s.

`Tools/MigrationSourcePrinter` renders a delta as C++ migration source, and `dbtool` gains two subcommands: `diff` (report only) and `generate <TITLE>` (write the migration), with `--empty` to scaffold a blank migration and `--up-to <TIMESTAMP>` to fold only migrations up to a point.

## Testing

| | Result |
|---|---|
| `clang-debug` (PEDANTIC + ASan + UBSan) | full suite green — 1402 cases |
| Databases | `sqlite3`, `mssql2022` (Docker), `postgres` (Docker 16.4) |
| `test_dbtool.py` | SUCCESS against `sqlite3` |
| `dbtool diff` | verified end to end against a live SQLite database |
| clang-tidy / clang-format | no new findings in the changed regions; `MigrationSourcePrinter.cpp` clean |

One portability fix is its own commit: `dbtool diff` rendered an index's column list with `std::views::join_with`, which is C++23 but not implemented in libc++, so the macOS CI leg failed to compile. Replaced with a small helper producing the identical string.

## Status — what still needs a look

- **Not yet reviewed for API shape.** `SchemaDiffResult` and friends are new public surface in `SqlMigration.hpp`; the naming and the owned-vs-view split deserve a second opinion before this leaves draft.
- **No tests for the diff itself.** The suite is green and `dbtool diff` was exercised by hand, but there is no automated coverage of `DiffAgainstLiveSchema` or `MigrationSourcePrinter` yet. That is the main gap.
- The modules (`LIGHTWEIGHT_BUILD_MODULES=ON`) and C++26-reflection configurations have **not** been built for this branch; `Lightweight.cppm` gained an entry, so per `AGENT.md` the modules build wants checking before merge.

- **Performance impact:** none on existing paths; the diff runs only when asked.
- **Risk:** medium — new public API in `SqlMigration.hpp` and two new `dbtool` subcommands. Nothing existing changes behaviour.

### DIFF
--- a/src/Lightweight/CMakeLists.txt
+++ b/src/Lightweight/CMakeLists.txt
@@ -197,6 +197,8 @@ target_compile_features(LightweightTools PUBLIC cxx_std_23)
 target_sources(LightweightTools PRIVATE
     Tools/CxxModelPrinter.cpp
     Tools/CxxModelPrinter.hpp
+    Tools/MigrationSourcePrinter.cpp
+    Tools/MigrationSourcePrinter.hpp
 )
 target_link_libraries(LightweightTools PUBLIC Lightweight)
 
--- a/src/Lightweight/Lightweight.cppm
+++ b/src/Lightweight/Lightweight.cppm
@@ -27,6 +27,7 @@ module;
 #include "SqlScopedLock.hpp"
 #include "ThreadSafeQueue.hpp"
 #include "Tools/CxxModelPrinter.hpp"
+#include "Tools/MigrationSourcePrinter.hpp"
 #include "Zip/ZipArchive.hpp"
 #include "Zip/ZipError.hpp"
 
--- a/src/Lightweight/SqlMigration.cpp
+++ b/src/Lightweight/SqlMigration.cpp
@@ -3056,8 +3056,12 @@ namespace
     /// `nvarchar (50)` must compare equal.
     [[nodiscard]] bool RenderedTypesEqual(std::string_view lhs, std::string_view rhs) noexcept
     {
-        auto const significant = [](char c) { return !std::isspace(static_cast<unsigned char>(c)); };
-        auto const fold = [](char c) { return static_cast<char>(std::toupper(static_cast<unsigned char>(c))); };
+        auto const significant = [](char c) {
+            return !std::isspace(static_cast<unsigned char>(c));
+        };
+        auto const fold = [](char c) {
+            return static_cast<char>(std::toupper(static_cast<unsigned char>(c)));
+        };
         return std::ranges::equal(lhs | std::views::filter(significant) | std::views::transform(fold),
                                   rhs | std::views::filter(significant) | std::views::transform(fold));
     }
--- a/src/Lightweight/SqlMigration.cpp
+++ b/src/Lightweight/SqlMigration.cpp
@@ -14,6 +14,7 @@
 
 #include <algorithm>
 #include <array>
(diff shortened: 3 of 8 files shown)


christianparpart:
@Yaraslaut this is still in draft?

PR #598, opened 2026-09-12
MS SQL Server: emit DATETIME2 instead of TIMESTAMP for Timestamp{} columns

### LINKED ISSUE #587: MS SQL Server: Timestamp{} emits TIMESTAMP, a rowversion synonym — column is binary and non-writable
Surfaced by the column-type coverage added in #579, where it is asserted as-is and labelled `KNOWN DEFECT` so it does not go unnoticed.

## Symptom

On MS SQL Server, a column declared through the migration query builder as

```cpp
migration.CreateTable("Example")
    .PrimaryKeyWithAutoIncrement("id")
    .RequiredColumn("timestampColumn", Timestamp {});
```

comes back out of the catalog as an 8-byte binary column, and `ddl2cpp` generates `Light::SqlDynamicBinary<8>` for it instead of `Light::SqlDateTime`. SQLite and PostgreSQL produce `Light::SqlDateTime` as expected.

## Cause

`SqlServerFormatter::ColumnType` emits the type name verbatim:

```cpp
// src/Lightweight/QueryFormatter/SqlServerFormatter.hpp:356
[](Timestamp const&) -> std::string { return "TIMESTAMP"; },
```

On SQL Server, `TIMESTAMP` is **not** a point in time. It is a deprecated synonym for `rowversion`: an 8-byte, server-generated, non-writable binary counter that changes on every update of the row. So the user asks for a timestamp and gets a read-only binary column.

## Impact

- A `Timestamp {}` column on SQL Server cannot be written to at all — inserts and updates targeting it are rejected by the server.
- Round-tripping such a schema through `ddl2cpp` produces a binary record member, so the C++ side inherits the wrong type too.
- The type is deprecated by Microsoft, so the spelling is on borrowed time regardless.

## Suggested fix

Emit `DATETIME2` for `Timestamp {}` in `SqlServerFormatter::ColumnType`. That is the natural counterpart to what the other two backends store, and it is what `DateTime {}` already maps to.

Worth deciding as part of the fix: whether existing SQL Server schemas created with the current spelling need a migration path, and whether the schema reader should keep recognising `rowversion`/`timestamp` columns as binary when reading third-party schemas (it should — that is a correct reading of what the column actually is).

## Verification

`src/tests/Ddl2CppColumnTypeTests.cpp` (from #579) has a dedicated `ColumnTypeTimestamp` case that currently branches on `SqlServerType::MICROSOFT_SQL` to expect `Light::SqlDynamicBinary<8>`. Once fixed, that branch should collapse to the single unconditional `Light::SqlDateTime` expectation.

Related: #191, #579

Yaraslaut:
On SQL Server, `TIMESTAMP` is a deprecated synonym for `rowversion` — an 8-byte, server-generated, non-writable binary counter, not a point in time. `SqlServerFormatter::ColumnType` emitted it verbatim for `Timestamp {}`, so the column came back out of the catalog as `binary(8)` and `ddl2cpp` generated `Light::SqlDynamicBinary<8>` instead of `Light::SqlDateTime`, and the column rejected every insert/update. SQLite and PostgreSQL already produced `Light::SqlDateTime` correctly — this was MSSQL-only.

## Changes

- `SqlServerFormatter::ColumnType` now emits `DATETIME2` for `Timestamp {}`, matching how the schema reader already maps `datetime2` back to `SqlColumnTypeDefinitions::DateTime`.
- The schema reader's handling of pre-existing `timestamp`/`rowversion` columns is intentionally unchanged — reading them as binary is a correct description of what such a column already is on disk.
- Updated `Ddl2CppColumnTypeTests.cpp`'s `ColumnTypeTimestamp` case to expect `Light::SqlDateTime` unconditionally (previously branched to expect `Light::SqlDynamicBinary<8>` on MSSQL), and `QueryBuilderTests.cpp`'s per-dialect `ColumnType` table to expect `DATETIME2` for SQL Server.

Fixes #587


christianparpart:
i'm slightly confused. maybe C++ `Timestamp` type should indeed be for SQL's `TIMESTAMP` whereas `DateTime{}` C++ type should be for SQL's `DATETIME` column type.

Yaraslaut:
Good question — and worth spelling out, since `TIMESTAMP` means something different on each backend.

The mapping isn't new here: `Timestamp {}` already produced `Light::SqlDateTime` on SQLite and PostgreSQL before this PR — `PostgreSqlFormatter::ColumnType` emits `TIMESTAMP` for it there too, and Postgres's `TIMESTAMP` *is* a real writable point-in-time type (it emits the same `TIMESTAMP` for `DateTime {}`, so the two are already synonyms on that backend). SQLite's type affinity makes the distinction moot as well.

SQL Server was the outlier, and not because it kept a stricter TIMESTAMP/DATETIME split — its `TIMESTAMP` isn't a temporal type at all, it's a deprecated synonym for `rowversion` (an 8-byte, server-generated, non-writable row-version counter). So `Timestamp {}` columns were silently binary and rejected every write (#587's actual symptom).

Given that, making `Timestamp {}` a real, writable datetime column on SQL Server (via `DATETIME2`) brings it in line with what it already meant everywhere else, rather than inventing new semantics. It does mean `Timestamp {}` and `DateTime {}` are effectively synonymous today on every backend — but that was already true on Postgres, so this PR doesn't introduce that overlap, it just removes SQL Server's divergence from it.

If we want `Timestamp {}` to mean something SQL-Server-specific like an actual `rowversion`/change-counter column in the future, that'd be a deliberate new feature (a distinct declared type, since it isn't a value you can write), not something to route through this bugfix.

